PR #1, opened 2021-04-04
Add checks for finished signage points

martomi:
Check that the full node doesn't skip any signage points. If it does, send a normal priority notification.


martomi:
- [ ] Add new notification to the README

PR #9, opened 2021-04-09
Adds script notifier

ajacobson:
I was looking for a way to trigger arbitrary actions based off of Chiadog events and I though that adding a generic script notifier would be a good approach.

I'm not currently doing anything with the script return codes to bubble up errors since it wasn't something I needed.

Let me know what you think.

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -14,3 +14,6 @@ notifier:
     enable: false
     api_token: 'dummy_token'
     user_key: 'dummy_key'
+  script:
+    enable: false
+    script_path: 'tests/test_script.sh'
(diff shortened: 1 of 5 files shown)


martomi:
I like the idea, thanks for the PR! 👍 

I assume your script is currently deciding what to do based on parsing the notification text? I might add unique identifiers to each notification to make this easier and to also be compatible with potential changes in the notification text.

ajacobson:
Yeah, I'm currently using the notification text to determine what to do.  Unique identifiers for different messages would be a great addition and clean up that logic substantially.

PR #13, opened 2021-04-13
TelegramNotifier

MiguelCarranza:
Simple [Telegram Bot](https://core.telegram.org/bots/api) Notifier

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -18,5 +18,6 @@ notifier:
     enable: false
     script_path: 'tests/test_script.sh'
   telegram:
+    enable: false
     bot_token: 'dummy_bot_token'
     chat_id: 'dummy_chat_id'
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -17,3 +17,6 @@ notifier:
   script:
     enable: false
     script_path: 'tests/test_script.sh'
+  telegram:
+    bot_token: 'dummy_bot_token'
+    chat_id: 'dummy_chat_id'
--- a/src/notifier/telegram_notifier.py
+++ b/src/notifier/telegram_notifier.py
@@ -22,11 +22,19 @@ class TelegramNotifier(Notifier):
         errors = False
         for event in events:
             if event.type == EventType.USER:
+                priority = ""
+
+                if event.priority == event.priority.HIGH:
+                    priority = "🚨"
+                elif event.priority == event.priority.NORMAL:
+                    priority = "⚠️"
+
                 request_body = json.dumps(
                     {
                         "chat_id": self.chat_id,
-                        "text": f"*{self._title_prefix} {event.service.name}*\n{event.message}",
+                        "text": f"{priority} *{self._title_prefix} {event.service.name}*\n{event.message}",
                         "parse_mode": "Markdown",
+                        "disable_notification": event.priority == event.priority.LOW,
                     }
                 )
                 conn = http.client.HTTPSConnection("api.telegram.org")
--- a/src/notifier/telegram_notifier.py
+++ b/src/notifier/telegram_notifier.py
@@ -1,0 +1,45 @@
+# std
+import http.client
+import logging
+import json
+from typing import List
+
+# project
+from . import Notifier, Event, EventType
+
+
+class TelegramNotifier(Notifier):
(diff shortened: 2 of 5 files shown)


config-example.yaml
```diff
@@ -14,3 +14,6 @@ notifier:
     enable: false
     api_token: 'dummy_token'
     user_key: 'dummy_key'
+  telegram:
+    bot_token: 'dummy_bot_token'
+    chat_id: 'dummy_chat_id'
```
martomi, line 19: Please add `enable: false` config key.

martomi (changes requested):
@MiguelCarranza looks good, thanks for the PR! Unfortunately I forgot to mention in CONTRIBUTION.md that PRs should be opened against the `dev` branch. I changed the target now - could you please rebase?

src/notifier/telegram_notifier.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def send_events_to_user(self, events: List[Event]) -> bool:
+        errors = False
+        for event in events:
+            if event.type == EventType.USER:
+                request_body = json.dumps(
+                    {
+                        "chat_id": self.chat_id,
+                        "text": f"*{self._title_prefix} {event.service.name}*\n{event.message}",
+                        "parse_mode": "Markdown",
```
greimela, line 29: We could disable the notification in case of a low event priority.
```suggestion
                        "parse_mode": "Markdown",
                        "disable_notification": event.priority == EventPriority.LOW,
```

src/notifier/telegram_notifier.py
```diff
@@ -0,0 +1,45 @@
...
+            logging.error(f"Invalid config.yaml. Missing key: {key}")
+
+    def send_events_to_user(self, events: List[Event]) -> bool:
+        errors = False
+        for event in events:
+            if event.type == EventType.USER:
+                request_body = json.dumps(
+                    {
+                        "chat_id": self.chat_id,
+                        "text": f"*{self._title_prefix} {event.service.name}*\n{event.message}",
```
greimela, line 28: I would add a 🚨 symbol in front of the text in case of a HIGH event priority, since Telegram does not support sending high priority messages itself.

tests/notifier/test_telegram_notifier.py
```diff
@@ -0,0 +1,37 @@
+# std
+import os
+import unittest
+
+# project
+from src.notifier import Event, EventType, EventPriority, EventService
+from src.notifier.telegram_notifier import TelegramNotifier
+
+
+class TestTelegramNotifier(unittest.TestCase):
```
greimela, line 10: I would add some tests for the HIGH and NORMAL priority cases, if they get handled differently.

greimela (commented):
Oh, I also created a TelegramNotifier yesterday evening, it looks quite similar 😄 

I will add some suggestions here, for things I added in my version.

src/notifier/telegram_notifier.py
martomi: Good idea!

config-example.yaml
MiguelCarranza: dang, I probably removed it when renaming the keys

MiguelCarranza:
> @MiguelCarranza looks good, thanks for the PR! Unfortunately I forgot to mention in CONTRIBUTION.md that PRs should be opened against the `dev` branch. I changed the target now - could you please rebase?

Comments addressed and rebased!

martomi:
Perfect thanks! Only nitpick I have left is to squash the commits to keep history cleaner, but I probably can also do it after rebasing a bit later today :-)

greimela:
@martomi You could just use squash as merge strategy, I like to use that in my projects as well.

martomi:
@greimela Thanks, I'll try that but I don't like that there's little control over the resulting commit message 😄

greimela:
AFAIK you can freely edit the resulting commit message as maintainer. Per default it uses the pull request title, but you can modify it before merging.

martomi:
Wow, that's very cool, thanks for the tip!

martomi:
@greimela did you also use Telegram Bot for your implementation? I'm updating the README now and trying to understand how do you ensure that the bot cannot be subscribed by a third person?

PR #15, opened 2021-04-13
Fix: Correctly handle logs created by Chia 1.0.4

### LINKED ISSUE #14: May need a fix for Chia 1.0.4
I'm getting some spurious 'HARVESTER: Your harvester appears to be offline! No events..' and I wondered if it's because I just upgraded to 1.0.4 (actually I appear to be on 1.0.5dev, so if you don't see this problem, it may be me.)

greimela:
The breaking change was https://github.com/Chia-Network/chia-blockchain/commit/7c9e7b44926ffb7ce4be1facd76cb1f09244258c.

Closes #14.


martomi:
Awesome, thanks for also updating the tests 👍 Shall we add a very small test for the backward compatibility?
Please squash the commit that makes it backward compatible with the fix to keep history cleaner.

greimela:
@martomi I extended the existing test case to test for backwards compatibility and squashed the commits ✅

martomi:
Perfect thanks!

PR #21, opened 2021-04-15
DiscordNotifier

kilbot:
Simple Discord notifier using [webhooks](https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks)


src/notifier/discord_notifier.py
```diff
@@ -0,0 +1,44 @@
...
+        for event in events:
+            if event.type == EventType.USER:
+                o = urllib.parse.urlparse(self.webhook_url)
+                conn = http.client.HTTPSConnection(o.netloc)
+                conn.request(
+                    "POST",
+                    o.path,
+                    urllib.parse.urlencode(
+                        {
+                            "username": f"chiadog on {socket.gethostname()}",
```
martomi, line 32: I like the idea of hostname but that is a bit more difficult to control. Let's use the title prefix which is configurable from the config.yaml file. The title prefix approach also works when starting multiple `chiadog` instances on the same machine that monitor a few remote harvesters. See how it's used in the Pushover notifier for example.

martomi: I also liked @greimela's [suggestion](https://github.com/martomi/chiadog/pull/13#discussion_r612297928) for 🚨 symbol for high priority events. What do you think, would that be valuable on Discord too?

See how we have it now on Telegram [for reference](https://github.com/martomi/chiadog/blob/main/src/notifier/telegram_notifier.py#L30), maybe you can use the same approach.

martomi (commented):
Looks good, have couple of improvement suggestions, let me know what you think. Also thanks for also updating the README!

README.md
```diff
@@ -76,6 +76,19 @@ Test your script is called correctly with:
 python3 -m unittest tests.notifier.test_script_notifier
 ```
 
+### Discord
+
+[Discord](https://pushover.net/)'s built in Webhooks are a super simple way to get notifications sent to a text channel in your server. Follow the instructions for *Making a Webhook* on the [Intro to Webhooks](https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks) page. Copy & paste the Webhook URL into your `config.yaml`, that's it!
```
martomi, line 81: URL is pointing to pushover website 😄

src/notifier/discord_notifier.py
kilbot: Yes, I think that is a good idea, I did think about something like that but was a bit lazy 😆 

Discord have [embeds](https://discordjs.guide/popular-topics/embeds.html#embed-preview) which can be used for rich content. I have a monitoring service on my little Raspberry Pi which puts out messages like this:

Screen Shot 2021-04-15 at 10 05 05 pm

martomi: Look cool, I see embeds are also supported through Webhooks API. Perhaps a bit more work, I'm also fine if you just follow the pattern we used for Telegram already 😄

kilbot: Embeds are probably overkill at the moment, they will just clog the chat. I have made the charges to follow the Telegram Notifier:

Screen Shot 2021-04-19 at 11 28 04 pm

martomi: Awesome! 👍

PR #23, opened 2021-04-19
Add SMTP Notifier

mikehw:
Adds an SMTP Notifier, using the example from AWS SES guide, but should work with any SMTP server.

The notification is put in the subject line since they are short, and the body of the message is left blank.


src/notifier/smtp_notifier.py
```diff
@@ -0,0 +1,73 @@
...
+
+                # Attach parts into message container.
+                # According to RFC 2046, the last part of a multipart message, in this case
+                # the HTML message, is best and preferred.
+                msg.attach(part1)
+                msg.attach(part2)
+
+                # Try to send the message.
+                try:
+                    server = smtplib.SMTP(self.host, self.port)
```
martomi, line 60: I've tested this with credentials for SMTP Relay from Mailjet, and at least with that service, the initialisation of the connection takes **a lot** of time. It takes 185 seconds to complete the test cases.

Did you observe similar delays? Might be better to initialise this connection once in the constructor and keep it alive so that we don't incur time cost of doing that on every notification.

I checked with the current logic, this will block the thread that reads the logs for about 30 seconds but then the parser should be able to catch up, so it might be fine for now.

martomi (commented):
Thanks for the contribution @mikehw !

src/notifier/smtp_notifier.py
```diff
@@ -0,0 +1,73 @@
...
+                    server.ehlo()
+                    server.starttls()
+                    # stmplib docs recommend calling ehlo() before & after starttls()
+                    server.ehlo()
+                    server.login(self.username_smtp, self.password_smtp)
+                    server.sendmail(self.sender, self.recipient, msg.as_string())
+                    server.close()
```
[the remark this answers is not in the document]
mikehw: Fixed, quit calls close and does some extra stuff

src/notifier/smtp_notifier.py
mikehw: I didn't take that long. AWS SES took ~5 seconds to send the 6 emails. Now trying to re-use the SMTP connection if the connection is still open.

martomi:
Looks good, thanks for addressing the feedback!

PR #25, opened 2021-04-21
Fix SMTP issue by always getting a new connection

mikehw:
The open SMTP connection would sometimes return a positive response to noop's but throw an error when trying to send an email. Reverting back to creating a new connection each time until someone finds a better way to persist the connection.


martomi:
Look good, thanks!

PR #35, opened 2021-04-24
Add daily notifications with farm status

martomi:
Add option to send out daily stats summary

    * New notification type DAILY_STATS
    * StatsManager that collects stats and sends out daily
    * Configurable time of day when the summary is sent out
    * Follow similar pattern to condition checkers for stat accumulators

SMTP Notifier: Send message in email body

    This is now necessary because daily summary notifications are longer and
    don't fit in the email subject.

Notifier: Add configurable notification categories

    * It's now possible to enable / disable daily summary notification
      individually for each integration. E.g. only receive daily summaries
      over E-mail and use Pushover only for alarms.
    * Same for wallet notifications which are only informatve

Config.YAML: Add comments to the example


martomi:
TODO:

- [ ] Update README

greimela:
Very nice!
Should we add some unit tests?

martomi:
Some additional test coverage would be great, indeed. Though I think I won't have a chance to add it for this release.

greimela:
Alright, lets do the unit tests separately!

martomi:
Rebased to cleanup fixup commits. Will go ahead with merge and let's do further testing on dev until release tomorrow.

PR #40, opened 2021-04-25
Remove wallet events and added coins parser / handlers

martomi:
This log has been removed from chia in the most recent versions so it
serves no purpose to have the logic here. I've submitted a PR to re-add
the log back, if that happens we can revert this change.

https://github.com/Chia-Network/chia-blockchain/pull/2720


martomi:
Probably better next time to implement it with RPC calls to the daemon if that's possible though.

Issue #7, opened 2021-04-05
Support for Windows


martomi:
I personally don't have Windows to test and am not familiar with how the chia environment there is setup but can provide guidance on the steps necessary to add such support.

Here are some jumping off points if someone wants to take over:

- We need a WIndowsLogConsumer that is analogous to the UNIX implementation in [FileLogConsumer](https://github.com/martomi/chiadog/blob/main/src/chia_log/log_consumer.py#L51). It needs to replace the [tail -F](https://github.com/martomi/chiadog/blob/main/src/chia_log/log_consumer.py#L69) command with the analogous command for PowerShell. 
   - Possibly something like `Get-Content <log-file> -Tail <#-of-rows> -Wait`
- Add `windows_log_consumer` config. See [here](https://github.com/martomi/chiadog/blob/main/config-example.yaml#L4) for an example. It needs to point to default path on Windows where `debug.log` is stored.
- The project is using [pathlib](https://docs.python.org/3/library/pathlib.html) for handling the file paths. I believe this is compatible with Windows paths but needs to be tested.

I suggest doing a quick test by directly modifying the [FileLogConsumer](https://github.com/martomi/chiadog/blob/main/src/chia_log/log_consumer.py#L51) and changing the path in the config to see if there are any other showstoppers to make this work on Windows. That will already be very valuable feedback.

skrustev:
I was able to successfully run it without any issues on Windows 10 with the latest Python 3.9.4 Windows distribution. 

The steps for running it are pretty much the same, only difference is that on Windows , the Python virtual environment is activated by running the `./.venv/Scripts/activate` script directly. 
 - If you are using VS Code, you can select the created virtual environment by pointing to the `./.venv/Scripts/python.exe` and use the Python terminal to run the ChiaDog.

What else I probably had to set differently is the `file_path` in the config, since on Windows its usually distributed under `C:/Users/<username>/.chia/mainnet/log/debug.log` for example.

stoatlemon:
Get the following error when trying to run on Windows 10 with Python 3.9.4 and modified debug.log file path. 
@skrustev any idea what I'm missing ? 

(venv) PS C:\Users\User\chiadog> python main.py --config config.yaml
[2021-04-20 16:41:07] [    INFO] --- Enabled file log consumer. (log_consumer.py:55)
[2021-04-20 16:41:07] [    INFO] --- Consuming log file from C:\Users\User\.chia\mainnet\log\debug.log (log_consumer.py:68)
[...]

PR #43, opened 2021-04-26
Add Windows support for file read and interrupt.

skrustev:
Should address any know issues so far for running on Windows described in #7

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -9,9 +9,6 @@ elseif (Get-Command python3 -errorAction SilentlyContinue) {
 # Create virtual environment
 Invoke-expression "$($command) -m venv .venv"
 
-# Create virtual environment
-Invoke-expression "$($command) -m venv .venv"
-
 # Activate the virtual environment
 ./.venv/Scripts/activate.ps1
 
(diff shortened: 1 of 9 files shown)


martomi:
Cool thanks! I'm going to push this into a minor release later today.

martomi:
I'd suggest to also add in a comment [here](https://github.com/martomi/chiadog/blob/main/config-example.yaml#L10) example path for windows, with the correct slashes // \

skrustev:
All types of paths actually work I think, since they are resolved by the pathlib. I've added a comment regarding being absolute path though.
Also a few easy to use scripts. They can be easily removed if not needed.

install.ps1
```diff
@@ -0,0 +1,27 @@
...
+}
+elseif (Get-Command python3 -errorAction SilentlyContinue) {
+    $command = "python3"
+}
+
+# Create virtual environment
+Invoke-expression "$($command) -m venv .venv"
+
+# Create virtual environment
+Invoke-expression "$($command) -m venv .venv"
```
martomi, line 13: This is duplicate?

PR #47, opened 2021-04-27
Added Windows Support for Network Log Consumer

pieterhelsen:
Uses `uname` over SSH to determine the Operating System that the remote platform is running on.
Switches from `tail` to PowerShell `Get-Content` when Windows is found.


README.md
```diff
@@ -254,13 +254,56 @@ ssh -i "~/.ssh/id_ed25519" <user>@<ip_address>
...
+
+```powershell
+$acl = Get-Acl C:\ProgramData\ssh\administrators_authorized_keys
+$acl.SetAccessRuleProtection($true, $false)
+$administratorsRule = New-Object system.security.accesscontrol.filesystemaccessrule("Administrators","FullControl","Allow")
+$systemRule = New-Object system.security.accesscontrol.filesystemaccessrule("SYSTEM","FullControl","Allow")
+$acl.SetAccessRule($administratorsRule)
+$acl.SetAccessRule($systemRule)
+$acl | Set-Acl
+```
```
martomi, line 304: Really like the extensive instructions. 👍  I've decided to make a wiki where we can put more detailed advanced setup instructions and other docs in the future to not overwhelm the README itself. It's also easier to edit by anyone. Please move this to the page I've created about [Monitoring Multiple Harvesters](https://github.com/martomi/chiadog/wiki/Monitoring-Multiple-Harvesters) (scroll to bottom).

src/chia_log/log_consumer.py
```diff
@@ -103,13 +105,33 @@ def stop(self):
...
+        stdin, stdout, stderr = self._ssh_client.exec_command(f"uname -a")
+        fout: str = stdout.readline().lower()
+        ferr: str = stderr.readline().lower()
+
+        if 'linux' in fout:
+            return 'linux'
+        elif 'darwin' in fout:
+            return 'macos'
+        elif 'not recognized' in ferr:
+            return 'windows'
```
martomi, line 129: Let's make `Enum` class for these platforms. Will be less error prone and can be reused in other places in the future. I would probably define it inside `util.py` file in `src` folder for now (but maybe you have better idea)

pieterhelsen: Should be OK now!

src/chia_log/log_consumer.py
```diff
@@ -103,13 +107,33 @@ def stop(self):
...
+        if 'linux' in fout:
+            return OS.LINUX
+        elif 'darwin' in fout:
+            return OS.MACOS
+        elif 'not recognized' in ferr:
+            return OS.WINDOWS
+        else:
+            logging.error(f"Found unsupported platform on remote host, assuming Linux and hope for the best.")
+
+        return 'linux'
```
martomi, line 135: This should also be `OS.LINUX`

PR #64, opened 2021-04-29
 Added Windows installation instructions

pieterhelsen:
Added detailed installation instructions for Windows.
Also updated the default README.md to cross-reference to the new Windows installation instructions.


WINDOWS.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+3. Create a virtual environment for `chiadog` to run in
+
+```
+# Create a new virtual environment in this folder called 'venv'
+python.exe -m venv venv
+
+# Activate the newly created virtual environment
+. ./venv/Scripts/activate
```
martomi, line 48: Should this not have the opposite slashes / \ ? (I don't know, asking 😄 )
Same about the other occurrences of that.

martomi (commented):
Great work, looks good!

WINDOWS.md
pieterhelsen: Good catch! That's what I get for typing from memory instead of copy/pasting.

martomi:
I think I can't exclude this from the PR myself. @pieterhelsen I think you did change permissions of install.sh and start.sh scripts by mistake? Can you drop these two permission changes from the PR?

pieterhelsen:
Looks like PyCharm on Windows automatically changes file permissions; updated.

martomi (approved):

PR #68, opened 2021-05-01
Improve cross-platform implementation

pieterhelsen:
Made the following changes to fix a Pathlib bug where the log file could not be found in the event of a Windows-run chiadog connecting to a remote Linux harvester:

- Renamed `_get_remote_platform()` to `get_host_info()` and made it static.

- Moved `get_host_info()` from class initializer to `create_log_consumer_from_config()`

- Replaced `Path` with either `PosixPurePath` or `WindowsPurePath` depending on  remote host OS info

- Added line to `.gitignore` to allow multi-remote testing

Also some minor updates `util.py` and log_consumer.py based on code formatter output

### DIFF
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -10,7 +10,7 @@ The latter has not been implemented yet. Feel free to add it.
 import logging
 import subprocess
 from abc import ABC, abstractmethod
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath, PurePath
 from threading import Thread
 from typing import List, Optional
 
@@ -83,19 +83,18 @@ class FileLogConsumer(LogConsumer):
 class NetworkLogConsumer(LogConsumer):
     """Consume logs over the network"""
 
-    def __init__(self, remote_log_path: Path, remote_user, remote_host):
+    def __init__(self, remote_log_path: PurePath, remote_user, remote_host, remote_platform=OS.LINUX):
         logging.info("Enabled network log consumer.")
         super().__init__()
 
         self._remote_user = remote_user
         self._remote_host = remote_host
         self._remote_log_path = remote_log_path
+        self._remote_platform = remote_platform
 
         self._ssh_client = paramiko.client.SSHClient()
         self._ssh_client.load_system_host_keys()
         self._ssh_client.connect(hostname=self._remote_host, username=self._remote_user)
-
-        self._remote_platform = self._get_remote_platform()
 
         # Start thread
         self._is_running = True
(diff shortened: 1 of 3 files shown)


src/chia_log/log_consumer.py
```diff
@@ -107,32 +106,42 @@ def stop(self):
...
-            return OS.LINUX
-        elif 'darwin' in fout:
-            return OS.MACOS
-        elif 'not recognized' in ferr:
-            return OS.WINDOWS
-        else:
-            logging.error(f"Found unsupported platform on remote host, assuming Linux and hope for the best.")
 
-        return OS.LINUX
+def get_host_info(host: str, user: str, path: str):
```
martomi, line 125: Please add return type annotation `-> Tuple[OS, PurePath]`

src/chia_log/log_consumer.py
```diff
@@ -107,32 +106,42 @@ def stop(self):
...
+    if "linux" in fout:
+        return OS.LINUX, PurePosixPath(path)
+    elif "darwin" in fout:
+        return OS.MACOS, PurePosixPath(path)
+    elif "not recognized" in ferr:
+        return OS.WINDOWS, PureWindowsPath(path)
+    else:
+        logging.error("Found unsupported platform on remote host, assuming Linux and hope for the best.")
+
+    return OS.LINUX
```
martomi, line 144: If you have specify the return type annotation as I suggest and run the [formatting, type-checking and linting](https://github.com/martomi/chiadog/blob/main/CONTRIBUTING.md#formatting-and-linting), it should automatically detect the wrong return here :-)

src/util.py
```diff
@@ -3,7 +3,6 @@
 
 class OS(Enum):
 
-    LINUX = 'LINUX'
-    MACOS = 'MACOS'
-    WINDOWS = 'WINDOWS'
-
+    LINUX = "LINUX"
+    MACOS = "MACOS"
+    WINDOWS = "WINDOWS"
```
martomi, line 8: Just to check: rre you using `black` for formatting as described in [CONTRIBUTION.md](https://github.com/martomi/chiadog/blob/main/CONTRIBUTING.md#formatting-and-linting)? Let me know if you need help to setup the parts there and if we need to update the manual. I will also have a look how to add some of these steps as automated PR checks soon.

src/chia_log/log_consumer.py
```diff
@@ -83,20 +83,19 @@ def _consume_loop(self):
 class NetworkLogConsumer(LogConsumer):
     """Consume logs over the network"""
 
-    def __init__(self, remote_log_path: Path, remote_user, remote_host):
+    def __init__(self, remote_log_path: PurePath, remote_user, remote_host, remote_platform=OS.LINUX):
```
martomi, line 86: Just noticed, we're missing type annotations for most arguments here, would be great if you can add them.

Let's not have default argument for `remote_platform` so that it's always a required argument. Otherwise it could be wrongly initialized by providing Windows path and not overriding the default platform.

In fact, to avoid that redundancy and potential mismatch you could actually determine the platform directly from the path class now: 

```isinstance(remote_log_path, PureWindowsPath)```

But I'll leave that choice up to you. Maybe it's better for future to have it separate and explicit.

martomi (commented):
Great use of pure paths, thanks for the PR!

src/util.py
pieterhelsen: Yes, this change was proposed by `black`. All in all that required very little setup.
I did notice that the documentation for running unit tests is not cross-platform. Setting the os environment variables as proposed by the documentation is not possible from a Windows command line, so I'll have to check if there is another (cross-platform) method to get those variables passed into the unit test

src/chia_log/log_consumer.py
pieterhelsen: I added the type annotations. I would prefer to leave it explicit for now, since PurePosixPath does not distinguish between MacOS and Linux. 

Currently, we don't specifically require the distinction either, but it might be handy in the future.

src/util.py
pieterhelsen: I was able to set the environment variables on Windows and updated the CONTRIBUTING.md manual to reflect those changes (will send a different PR for this)

martomi (approved):

PR #71, opened 2021-05-02
Add MQTT Notifier

pieterhelsen:
Added MQTT Notifier and Unit Tests. 
The notifier takes an IP address or hostname, port number and MQTT topic at minimum and supports optional username/password authentication. Retain and QoS parameters can also be passed as config variables.

When called, the notifier will push a message to the provided topic. 
The message will be formatted as:
`{"type": event.type.name, "prio": event.priority.name, "msg": event.message}`

with `type` as:

- USER
- KEEPALIVE
- DAILY_STATS

and `priority` as:

- LOW
- NORMAL
- HIGH

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 paramiko==2.7.2
 python-dateutil~=2.8.1
 PyYAML==5.4
+paho-mqtt==1.5.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [0.4.1] - 2021-04-27
+
+### Added
+
+- Daily report for number of occasions where plot seeking time is above 5 and 15 seconds.
+  See [this issue](https://github.com/martomi/chiadog/issues/49) to understand why you really want response time < 5
+  seconds and not 30 seconds. You're going to get a notification for anything over 20 seconds which is in the critical
+  range.
+- Windows: Experimental support (thanks [@skrustev](https://github.com/skrustev)) - not yet a one-click easy setup - it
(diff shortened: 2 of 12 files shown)


greimela:
I am currently working on a generic HTTP stats reporter, which sends statistics to a HTTP endpoint in regular intervals (e.g. every 5 mins).
It uses the same `stat_accumulators` as the `daily_stats` reporter, only in a structured format.
I suspect this would be helpful in your Home Assistant use case.
Maybe we can combine it?

The event could look something like this:
```
{
  'timestamp': 1620117180, 
  'found_proofs_total': 0, 
  'search_time_average': 0.32, 
  'search_time_over_5s_count': 0, 
  'search_time_over_15s_count': 0, 
  'current_plot_count': 440, 
  'eligible_plots_average': 1.17, 
  'signage_point_count': 25, 
  'skipped_signage_point_count': 0
}
```

martomi:
@pieterhelsen we need to cleanup the commits on that branch. I think you've used merging of dev instead of rebasing, while I'm using rebasing and that probably is why everything is shown duplicate here. I recommend cherry-picking the relevant commits onto a new branch and using rebase on dev in the future.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 paramiko==2.7.2
 python-dateutil~=2.8.1
-PyYAML==5.4
\ No newline at end of file
+PyYAML==5.4
+paho-mqtt==1.5.1
```
martomi, line 4: I think we don't want to add this as a requirement. It's a feature only 1% or less of users will use.
Maybe we can use [this method of optional imports](https://stackoverflow.com/a/563060) and specify additional requirements required for this integration in the `INTEGRATIONS.md`.

pieterhelsen:
Closed in favor of #94

PR #73, opened 2021-05-03
Add total searches to daily stats

pieterhelsen:
Added overall total number of searches. Also added percentage of total searches for 5s/15 measurements

### DIFF
--- a/src/chia_log/handlers/daily_stats/stat_accumulators/search_time_stats.py
+++ b/src/chia_log/handlers/daily_stats/stat_accumulators/search_time_stats.py
@@ -29,9 +29,17 @@ class SearchTimeStats(HarvesterActivityConsumer, StatAccumulator):
             self._over_15_seconds += 1
 
     def get_summary(self) -> str:
+
+        pct_over_5seconds: float = 0
+        pct_over_15seconds: float = 0
+
+        if self._num_measurements > 0:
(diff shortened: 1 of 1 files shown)


pieterhelsen:
changes from running `black` and `flake8` have been reverted as they are already solved in the mqtt-notifier branch

martomi:
I see the mqtt branch is a bit of a mess now 😄 Every individual branch going to dev should be formatted and tested by itself and not depend on any other branch. Once it's merged, the way to go is to rebase other branches on top of dev. Something like this https://gist.github.com/markreid/12e7c2203916b93d23c27a263f6091a0

Do you have an example of how this looks like on phone format? Does it fit the line?

pieterhelsen:
On discord:

Laptop: 
image

Mobile:
Screenshot_20210508-110140_Discord

martomi:
Awesome, thanks for visuals! 

Nitpicking here but I think it would be nice to make it a bit less verbose and compact like this:

```
Search:
  - average: 0.24s over 420 searches
  - over 5s: 20 (4,7%)
  - over 15s: 1 (0.2%) 
```
Do you think it's still readable or better keep the verbose way? In any case we can remove the "average" repetition.

pieterhelsen:
Closed in favor of #95

PR #74, opened 2021-05-03
Bring back added coins / reward notifications!

martomi:
As of Chia 1.1.3, the log for added coins is active again so we can
restore that feature.

### DIFF
--- a/src/notifier/__init__.py
+++ b/src/notifier/__init__.py
@@ -40,7 +40,8 @@ class EventService(Enum):
     HARVESTER = 0
     FARMER = 1
     FULL_NODE = 2
-    DAILY = 3
+    WALLET = 3
+    DAILY = 4
 
 
 @dataclass
@@ -69,6 +70,8 @@ class Notifier(ABC):
             if config["daily_stats"]:
                 self._notification_types.append(EventType.DAILY_STATS)
                 self._notification_services.append(EventService.DAILY)
+            if config["wallet_events"]:
+                self._notification_services.append(EventService.WALLET)
         except KeyError as key:
             logging.error(f"Invalid config.yaml. Missing key: {key}")
 
--- a/tests/chia_log/handlers/test_wallet_added_coin_handler.py
+++ b/tests/chia_log/handlers/test_wallet_added_coin_handler.py
@@ -1,0 +1,28 @@
+# std
+import unittest
+from pathlib import Path
+
+# project
+from src.chia_log.handlers.wallet_added_coin_handler import WalletAddedCoinHandler
+from src.notifier import EventType, EventService, EventPriority
+
+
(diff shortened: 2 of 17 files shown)


src/notifier/__init__.py
```diff
@@ -40,7 +40,8 @@ class EventService(Enum):
     HARVESTER = 0
     FARMER = 1
     FULL_NODE = 2
-    DAILY = 3
+    WALLET = 3
```
pieterhelsen, line 43: Probably not really an issue at this time, but it might be better to append rather than change the values of an enum, considering this may be a breaking change for integrations at some point.

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -0,0 +1,28 @@
...
+from src.notifier import EventType, EventService, EventPriority
+
+
+class TestWalledAddedCoinHandler(unittest.TestCase):
+    def setUp(self) -> None:
+        self.handler = WalletAddedCoinHandler()
+        self.example_logs_path = Path(__file__).resolve().parents[1] / "logs/wallet_added_coin"
+
+    def testNominal(self):
+        with open(self.example_logs_path / "nominal.txt") as f:
```
pieterhelsen, line 16: Encoding needs to specifically set to UTF-8 or unit tests will fail under Windows

tests/chia_log/parsers/test_wallet_added_coin_parser.py
```diff
@@ -0,0 +1,27 @@
...
+
+# project
+from src.chia_log.parsers.wallet_added_coin_parser import WalletAddedCoinParser
+
+
+class TestWalletAddedCoinParser(unittest.TestCase):
+    def setUp(self) -> None:
+        self.parser = WalletAddedCoinParser()
+        self.example_logs_path = Path(__file__).resolve().parents[1] / "logs/wallet_added_coin"
+        with open(self.example_logs_path / "nominal.txt") as f:
```
pieterhelsen, line 13: Encoding needs to be specifically set to UTF-8 or unit tests will fail under Windows

pieterhelsen (approved):
Looks good! 

I've updated the mqtt-notifier branch to reflect the `wallet_events` changes in the unit test and config example.

src/notifier/__init__.py
martomi: Good catch! The order only got shuffled like that because I did the features in reverse order the first time.

greimela:
I think I lost track of the state of this PR.
Can I contribute anything to get it merged?

martomi:
Missing one approval to get merged basically 😄 @pieterhelsen approved before but I have configured to reset approvals after the PR is modified.

Issue #69, opened 2021-05-01
Add custom frequency config option for daily stats report & add manual trigger


centrd:
Currently, chiadog can send a daily stats report, which is awesome. 
Request: 

1. Add a config field to set the frequency of this report, e.g. every 5 hours, every 1 hour, etc. This is especially helpful while troubleshooting issues with the farmer.

2. Add the ability to trigger a stats notifications manually. Again, this is helpful for troubleshooting. It would be extra nice if this were possible to call via CLI, a script, or cron.

bstock84:
+1 to this request. 

In addition, I'd love to be able to get a notification when a proof is found, especially since chia's debug.log only goes back 8 files, and it rotates every ~20MB (which is my case is every 40 minutes), which means I only end up with 4-5 hours of logs.

pieterhelsen:
> +1 to this request.
> 
> In addition, I'd love to be able to get a notification when a proof is found, especially since chia's debug.log only goes back 8 files, and it rotates every ~20MB (which is my case is every 40 minutes), which means I only end up with 4-5 hours of logs.

There is a pull request ready for this functionality. It was briefly disabled due to Chia removing the log line from the debug log. 
https://github.com/martomi/chiadog/pull/74

pieterhelsen:
> Currently, chiadog can send a daily stats report, which is awesome.
> Request:
> 
> 1. Add a config field to set the frequency of this report, e.g. every 5 hours, every 1 hour, etc. This is especially helpful while troubleshooting issues with the farmer.
> 2. Add the ability to trigger a stats notifications manually. Again, this is helpful for troubleshooting. It would be extra nice if this were possible to call via CLI, a script, or cron.

Regarding 1) so basically you'd like to have a user-defined interval for ~~daily~~ stats? Config would look something like this then:

```
daily_stats:
  enable: true
  start_at: 21:15
  interval: 3600 # interval in seconds or preset: [hourly, daily, weekly]
```

Regarding 2) I've been thinking of a way to implement this. Since Chiadog keeps its state in memory, I can think of two possible ways:

- Write state to shared database and have a command `stats.py --config config.yaml` read from that database, send a notification and exit
- Implement a keyboard event listener using `keyboard` and fire off a manual trigger when you press a specific key/combination (like ctrl+t) from the existing main.py process. 

I'm leaning towards the second implementation... If anyone can think of any other ways, please elaborate on them here.

greimela:
Regarding 2: 
[...]

PR #76, opened 2021-05-04
Add HTTP stat reporter

greimela:
This PR adds a stat reporter that reports accumulated stats to a HTTP endpoint.
It reuses the same stat_accumulators as the daily stats reporter, in a structured format.

The resulting event currently looks like this:
```
{
  'timestamp': 1620117180, 
  'found_proofs_total': 1, 
  'search_time_average': 0.32, 
  'search_time_over_5s_count': 3, 
  'search_time_over_15s_count': 0, 
  'current_plot_count': 1234, 
  'eligible_plots_average': 10.17, 
  'signage_point_count': 25, 
  'skipped_signage_point_count': 0
}
```

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -27,9 +27,10 @@ keep_alive_monitor:
 
 # Enable this and you'll receive a daily summary notification
 # on your farm performance at the specified time of the day.
-daily_stats:
-  enable: true
-  time_of_day: 21
+stats:
+  daily_stats:
+    enable: true
+    time_of_day: 21
 
 # We support a lot of notifiers, please check the README for more
 # information. You can delete the sections which you aren't using.
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -31,6 +31,10 @@ stats:
   daily_stats:
     enable: true
     time_of_day: 21
+  http:
+    enable: true
+    hostname: 'example.com'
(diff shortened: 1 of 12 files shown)


martomi:
The implementation fits fairly nicely in the current architecture. However, I personally don't see much value in sending that information to external HTTP API, especially after implementing https://github.com/martomi/chiadog/issues/69 which essentially brings the same information as notifications at the frequency the user wants.

I can see some value in seeing aggregated historical data charts but it's very likely that such an external HTTP aggregator will need to collect more data to grow and further differentiate from the existing notifications. That expansion will be in direct conflict with the idea of keeping this tool simple & minimal. 

For that reason, I'd say a fork or a separate tool (which can reuse the parsers from here as per the MIT license) would be the correct long-term choice for this development.

greimela:
I understand your valid concerns and agree that chiadog should stay as minimal as possible!
Therefore, I will close this PR for now.

Still, there are some aspects that fit very nicely in the context of the chiadog as a notification tool, e.g. a central monitoring of a decentralized farm with multiple harvesters.
This use case is currently not well supported, since each device has its own chiadog process, inspecting its own log.

But as you said, a new tool could reuse relevant parts from this project and build on it!
Thanks again for making this tool open source! 😉

PR #81, opened 2021-05-05
Hotfix WindowsPath

### LINKED ISSUE #80: Windows - Dev branch - WindowsPath not iterable
**
I got this in the dev branch on line 76 of log_consumer.py

python popen argument of type 'WindowsPath' is not iterable

**Environment:**
 - OS: Windows 10
 - Python version: 3.9.5

** 
I changed the code on line 68 to:
        expanded_user_log_path = str(self._log_path.expanduser())

which fixes it.  I tried to push a branch with that small change for you guys to review, but I have no permission to push

pieterhelsen:
Expanded path needed to be passed to Popen as a string.

Closes #80

### DIFF
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -65,7 +65,7 @@ class FileLogConsumer(LogConsumer):
         self._is_running = False
 
     def _consume_loop(self):
-        expanded_user_log_path = self._log_path.expanduser()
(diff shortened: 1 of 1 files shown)


martomi (approved):
Thanks! Still don't fully understand why path-like object doesn't work on Windows because subprocess.Popen does support it as an argument. But seems harmless to add explicit conversion 😄

PR #82, opened 2021-05-05
Add startup notification

pieterhelsen:
Added an event that notifies all initialized notifiers that chiadog has started

### DIFF
--- a/src/notifier/__init__.py
+++ b/src/notifier/__init__.py
@@ -30,6 +30,7 @@ class EventType(Enum):
     KEEPALIVE = 0
     USER = 1
     DAILY_STATS = 2
+    STARTUP = 3
 
 
 class EventService(Enum):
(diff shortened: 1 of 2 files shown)


martomi (changes requested):
Thanks for having a go at implementing that feature request.

There's couple things I don't like:
- We need to add new notification type and service for a very narrow feature
- It's not configurable, many people might not like it every time

 I'd propose a little different idea that keeps things simpler and doesn't require new notification type and service:

1) Add config option here that also sends out daily notification on startup: https://github.com/martomi/chiadog/blob/main/config-example.yaml#L32

2) If config is enabled, send out daily notification already from within the constructor here: https://github.com/martomi/chiadog/blob/main/src/chia_log/handlers/daily_stats/stats_manager.py 

What you think?

pieterhelsen:
OK. Makes sense!

pieterhelsen:
Closed in favor of #185

PR #86, opened 2021-05-05
Install.sh: Add python check

martomi:
Added Python check to verify python version is 3.7 or above before proceeding with installation


martomi:
@pieterhelsen Before merging, please select Squash and merge method and add prefix to the title of the commit so it's `Install.sh: Add python check`, otherwise it's too ambigious where that check is added.

martomi:
Nevermind this PR, I figured how to merge it without unnecessary re-approval 😄

PR #90, opened 2021-05-06
Add Windows Powershell launch script

fiveangle:


martomi:
Thanks! We actually [used to have this](https://github.com/martomi/chiadog/blob/89443132b1a24360e6def11b973df23209d88a49/start.ps1) but removed it in favour to explicitly running all commands. It seemed like it's not a good approach to ask users to disable safety checks in their PowerShell to run this.

PR #91, opened 2021-05-06
Add automated launching for Windows

fiveangle:


martomi:
This is only required in the context of this change https://github.com/martomi/chiadog/pull/90 right? 
In that case we can close it for the reasons mentioned there.

Alternatively, we could add advanced section at the very bottom and mention how that script can be created for anyone that wants this explicitly.

fiveangle:
Ah, I see your point.  Let me change it then... there is a relatively simple mechanism that allows alleviating the restriction only for the context of the script being launched and since the script is being launched is by the users action just as archaically assembling the python command-line, it is explicitly no difference from a security standpoint, and doesn't expose the user to potential system-wide scripting attacks.

I see you closed the #90 without the opportunity to address your (very real) security concern, so I will adjust this PR to accommodate the feature.  It will just take a bit more testing which I don't have time immediately but will get to it shortly, and if I find security model has changed with WIndows to not allow my idea, I will modify this PR with the script creation instructions as you suggest.

Thanks for considering the security of chiadog users !

fiveangle:
This should be working now !

image

fiveangle:
Reverted my env customization with the generic text placeholder, and fixed minor typo.  G2G now

martomi:
@pieterhelsen what do you think? I think we can include this but I'd prefer that we move it into a `scripts` directory and maybe even separate folders for `unix` and `windows`. Otherwise the root directory becomes somewhat unorganised with `start.sh` `start.cmd` `powershell.ps1` etc

fiveangle:
I just put them there because `start.sh` was there.  It definitely could get very messy very quickly so yeah, it makes sense to organize now.  The idea of having a unifed "start.*" launch script at the root for all the different platforms, with any helper scripts in a ./scripts/[platform]/ directory would probably be nice because it keeps the user experience cleaner staying in the root directory for the most part if using the cmdline, But yeah, good thinking.  Let me know if you want any direct changes from me !

powershell.ps1
```diff
@@ -0,0 +1,3 @@
+cd C:\Users\[YOUR-USER]\chiadog
```
martomi, line 1: Can we automate this as well? See https://stackoverflow.com/a/2085755

martomi:
@fiveangle I'd prefer to merge it after we have that idea implemented to some extend. Could already be good enough to just move the powershell script to scripts folder and rename it to `start.ps1` so that we only have user-facing scripts in the root and no scripts with generic non-descriptive names like `powershell.ps1`.

powershell.ps1
fiveangle: It's certainly possible.  In reality, the user can place the directory anywhere on-disk.  To be accurate, it should be changed to "cd <path>" but according to your instructions, you used this placeholder notation to correlate with your "quick guide" instructions.  It's rather easy to get the username for every platform, but I assume you hadn't automated that since a savvy user would immediately read between the lines to know "put your desired path here".

If you want to hard-code chiadog to the metaphorical "~/chiadog" on every platform, then it probably makes sense to merge this as is, then do an all-encompassing commit to do an automated home-dir-path on every platform.  I personally think it goes a bit too far, but hey... it's not my project :)   If you want me to automate hard-coding the homedir to `%userprofile%/chiadog` so that you can handle every other platform shortly after, that's fine.  Lemme know !

fiveangle: with just forcing the referenced path to the platform-specific directory and keying off the CWD for the launch stub (`start.cmd`), we actually don't need the user to hand-jam the powershell launch script 👍 

I've created the "scripts" dir for platform-specific stuff and made the discussed changes

fiveangle:
check it out and let me know what you think

martomi:
Thanks @fiveangle, looks good! Have you tested it with this folder structure? Is the powershell script running in the working directory of the `start.cmd` script, or how does it find the `main.py` from the scripts folder?

fiveangle:
yes, both cmdline and launching via shortcut. It runs from the context of the calling `start.cmd` directory, which allows use of relative paths, eliminating the need for user to edit the powershell script.

I just noticed an innocuous issue where I had used powershell comment notation in the cmd so fixed that (and added a small style tweak to the powershell script to follow standard powershell comment-based help context style.  should be good to go!

powershell.ps1
martomi: Perfect, thanks for thinking that through and resolving it in a meaningful way. I didn't consider the full context with the initial suggestion but you understood the idea. 👍

martomi:
@pieterhelsen another failure case for the auto-black 😄 This won't work for PRs from forks, I suggest to drop the idea of automated commits and just perform a simple check, what you think?

pieterhelsen:
> @pieterhelsen another failure case for the auto-black 😄 This won't work for PRs from forks, I suggest to drop the idea of automated commits and just perform a simple check, what you think?

Yup, looks like that is the simplest at this time. Will revert tonight.

Issue #89, opened 2021-05-06
FEATURE REQUEST: ability to use custom SSH port


echoparkbaby:
multiple harvesters. Am not able to connect because chiadog uses port 22 for SSH. I use a custom port. Would be nice to be able to change this in the prefs or yaml.

```
python3 main.py --config config-harvester-1.yaml
paramiko.ssh_exception.NoValidConnectionsError: [Errno None] Unable to connect to port 22 on 192.168.1.224
```

pieterhelsen:
Closed by #101

PR #101, opened 2021-05-10
Custom SSH port - issue 89

turekjiri:
Custom SSH port for remote logs can now be specified in config.yaml (issue #89)

### DIFF
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -168,23 +168,26 @@ def create_log_consumer_from_config(config: dict) -> Optional[LogConsumer]:
 
     if enabled_consumer == "network_log_consumer":
         if not check_keys(
-            required_keys=["remote_file_path", "remote_host", "remote_port", "remote_user"],
+            required_keys=["remote_file_path", "remote_host", "remote_user"],
             config=enabled_consumer_config,
         ):
             return None
+
+        # default SSH Port : 22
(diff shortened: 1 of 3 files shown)


src/chia_log/log_consumer.py
```diff
@@ -164,20 +168,23 @@ def create_log_consumer_from_config(config: dict) -> Optional[LogConsumer]:
 
     if enabled_consumer == "network_log_consumer":
         if not check_keys(
-            required_keys=["remote_file_path", "remote_host", "remote_user"], config=enabled_consumer_config
+            required_keys=["remote_file_path", "remote_host", "remote_port", "remote_user"],
```
martomi, line 171: Thanks for the contribution!

Perhaps worth making the port optional so we have a bit better backward compatibility with existing config files, otherwise looks good to me! 

@pieterhelsen tagging you here since you had that issue assigned to yourself, please check it out!

pieterhelsen: Agree, would make the port optional in the config and set it explicitly to 22 if the config is not present.

pieterhelsen (commented):
Thanks a bunch for your work on this! If you can address that minor change in the config-example.yaml, we're good to go 💯

turekjiri:
> Thanks a bunch for your work on this! If you can address that minor change in the config-example.yaml, we're good to go 💯

Done 👍

martomi:
@turekjiri Thanks for quickly addressing all feedback!

martomi (approved):

PR #109, opened 2021-05-13
Improve skipped signage points

martomi:
* Dowgrade log output about out-of-order events to debug log level to
  reduce questions and concerns about this (expected) phenomenon
* Add clarification to the comment explaining when it happens
* Add validity flag in the return that allows the users of the function
  to clear their state from wrong previous SP records. Thus preventing
  incorrect signage point skip notifications.


pieterhelsen (approved):
As discussed, we'll preferably add a test case and test data specifically for this at some point. But for now, good to go!

PR #118, opened 2021-05-14
Add Docker instructions

jouir:
I'm running Chia on Docker using the [official Chia Docker container](https://github.com/Chia-Network/chia-docker). I wanted to integrate `chiadog` to my Docker environment so this commit brings a Dockerfile and instructions on how to setup `chiadog` using Docker.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,0 +1,59 @@
+## Installation on Docker
+
+### Clone the repository
+
+```
+git clone https://github.com/martomi/chiadog.git
+cd chiadog
+```
+
+### Build the Docker image
+
+#### Latest version
+
+```
+docker image build -t chiadog:latest .
+```
+
+#### Specific version
+
+```
+docker image build -t chiadog:0.5.0 --build-arg BRANCH=v0.5.0 .
+```
+
+### Copy the example config file
+
+```
+cp config-example.yaml config.yaml
+```
+
+### Edit the configuration
+
+Open up `config.yaml` in your editor and configure it to your preferences.
+
+## Monitoring a local harvester / farmer
+
+### Start the watchdog
+
+```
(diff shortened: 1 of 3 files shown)


DOCKER.md
```diff
@@ -0,0 +1,59 @@
...
+#### Specific version
+
+```
+docker image build -t chiadog:latest .
+```
+
+#### Latest version
+
+```
+docker image build -t chiadog:0.5.0 --build-arg BRANCH=v0.5.0 .
```
martomi, line 21: Is there a way we can have this such that updates with every new version of chiadog are not necessary? E.g. using main branch and `latest` tag 🤔 Perhaps fetch version from a VERSION file in the root of the repository that we can maintain at central place?

martomi (commented):
Thanks for the contribution!

Has anyone else already tried running this, any feedback? I haven't had a chance yet.

skweee:
Great start! I agree that it would be good to have a Dockerfile in this repo. I have a couple of improvement wishes though, which you can also find in my pull request for the repo by @ajacobson (that's the one I'm using right now, but I would gladly switch):
https://github.com/skweee/chiadog-docker/blob/feature/add-branch-argument/Dockerfile

* Please also make the repository configurable to support builds from forks while making the default the original repo.
* To minimize the image size, you only need to make a shallow clone of the repo ("git clone --depth 1 [...]").
* I would be great if you could make the time zone configurable (e.g. using the TZ env variable as in the example above - also note the entrypoint.sh). Otherwise people will have to set the time of day for their daily stats using UTC.

Thanks for the contribution!

DOCKER.md
jouir: Build instructions for a specific version and the latest version were inverted. By default, if BRANCH argument is not supplied, the main branch is used so the image will use the latest version at the time of the build.

jouir:
I didn't know there was a separate repository doing exactly what I was looking for. No need to re-invent the wheel, I'm closing this pull-request. Thanks for sharing!

PR #119, opened 2021-05-14
Fix wallet events

skweee:
* amount in debug.log is no longer quoted
* check by sending yourself some mojos using https://faucet.chia.net/


bhy:
You are right that the amount is no longer quoted thus the parser needs to be fixed. However I think the new regex should accept both the quoted and non-quoted version for the sake of backward compatibility.

skweee:
Thanks for your feedback. Initially I also thought about providing backwards compatibility in this case. However, that would mean supporting older chia versions that still have bugs which compromise the overall health of the network (e.g. negative spends). I therefore think it is better to drop backwards compatibility and point people to upgrading their chia installation.

If there is general concensus about the need for backwards compatibility, I'd be happy to include it in this PR.

bhy:
@skweee that sounds quite reasonable! Thank you for the work!

martomi (approved):
Thanks for the fix! I agree with you on that one, no need for backward compatibility here.

PR #125, opened 2021-05-15
windows wsl installation instructions added

ougni:
Turns out installing on WSL-Windows works just fine as long as the path for the Cha log is specified correctly. Added some instructions for installation using WSL-Ubuntu on Windows.

### DIFF
--- a/WINDOWS.md
+++ b/WINDOWS.md
@@ -28,7 +28,7 @@ are available under `C:\Users\[YOUR-USER]\.chia\mainnet\log\debug.log`.
 2. Find the line that reads `log_level: DEBUG` (under the `farmer` section) and change this to `log_level: INFO`
 3. Restart the GUI or run `chia start --restart farmer` from the command line
 
-## 1.1 Installation (Powershell)
+## 1.1 Powershell(Native) Installation
 
 _For updating from previous version, see section below._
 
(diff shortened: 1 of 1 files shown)


chadmccune:
Does this solve the false offline notification problem?

ougni:
Was not aware of that problem. But the short answer is NO. 
This just proposes you run the chia-dog on WSL-Ubuntu instead of native Windows (with simplers installation process).

chadmccune:
I thought it might because I dont think the problem exists on linux as far as i'm aware

WINDOWS.md
```diff
@@ -52,12 +52,50 @@ python.exe -m venv venv
...
+    ```
+
+    4.2 configure it to your preferences for notification services(same as OSX/Ubuntu)
+
+---
+
+**Note**: if you get an error saying 'The term 'python.exe' is not recognized' or 'Python was not found', change
+`python.exe` in the last line of the above code block to `py.exe` or `python3.exe`.
+If none of these work, it is likely that Python was not installed on your system. Refer to
+[this website](https://www.python.org/downloads/windows/) in order to download the latest version.
```
martomi, line 96: I think that note should either be replicated or be higher up closer to the first section for installing natively on windows. Otherwise most people following the instructions with Powershell won't see it.

martomi:
My guess is also that it will solve the mentioned above problem if it reuses the `tail -F` command.

chadmccune:
I was hoping so, but so far in my testing it does not because I'm actually
getting the problem more often now....I think its a way the logs roll
over in windows, problem probably wouldn't exist if I was also running chia
on wsl2

WINDOWS.md
```diff
@@ -77,9 +77,47 @@ deactivate
...
+```
+
+4.  Open up `config.yaml` in your editor
+
+    4.1 Set chia log location:
+
+    ```yaml
+    chia_logs:
+      file_log_consumer:
+      enable: true
```
pieterhelsen, line 113: Missing indentation

WINDOWS.md
```diff
@@ -77,9 +77,47 @@ deactivate
...
 
-6. Open up the newly created `config.yaml` in your favorite text editor and configure it 
+6. Open up the newly created `config.yaml` in your favorite text editor and configure it
    to your preferences.
 
+## 1.2 WSL(Ubuntu) Installation
+
+Reminder: `INFO` level log in Chia must be enabled (`chia configure -log-level=INFO`)
+
+1. Open `WSL-Ubuntu`
```
pieterhelsen, line 87: I would explain briefly what WSL is and provide a link to an article detailing its installation.

WINDOWS.md
```diff
@@ -77,9 +77,47 @@ deactivate
...
+    4.1 Set chia log location:
+
+    ```yaml
+    chia_logs:
+      file_log_consumer:
+      enable: true
+      file_path: "/mnt/c/Users/<user-name>/.chia/mainnet/log/debug.log"
+    ```
+
+    4.2 configure it to your preferences for notification services(same as OSX/Ubuntu)
```
pieterhelsen, line 117: Nitpicking: missing space between 'services' and '(same...' 😃

pieterhelsen (changes requested):
Overall good addition to the documentation. I would add a brief explanation of WSL, so people are not confused when deciding between both options.

pieterhelsen (approved):
Looks good!

martomi (approved):

PR #131, opened 2021-05-16
Add --version switch

### LINKED ISSUE #128: Add switch to display version number
add --version switch 
or have -h display the version when called

pieterhelsen:
Initial implementation of a version switch as requested by #128 .

Not sure about the implementation yet, as it changes `--config` from a required to an optional variable. 
I also added a logline to the startup sequence of Chiadog that mentions the version. 
Thoughts are appreciated!

This PR closes #128

### DIFF
--- a/main.py
+++ b/main.py
@@ -2,10 +2,15 @@
 import argparse
 import logging
 import signal
+import subprocess
+import sys
 import time
+from argparse import Namespace, ArgumentParser
 from pathlib import Path
 
 # project
+from typing import Tuple
+
 from src.chia_log.handlers.daily_stats.stats_manager import StatsManager
 from src.chia_log.log_consumer import create_log_consumer_from_config
 from src.chia_log.log_handler import LogHandler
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -2,10 +2,15 @@
...
 import signal
+import subprocess
+import sys
 import time
+from argparse import Namespace, ArgumentParser
 from pathlib import Path
 
 # project
+from typing import Tuple
+
```
martomi, line 13: this should be above, not under project includes

main.py
```diff
@@ -87,3 +92,23 @@ def interrupt(signal_number, frame):
...
+    argparse, args = parse_arguments()
+
+    if args.config:
+        conf = Config(Path(args.config))
+        init(conf)
+    elif args.version:
+        print(version())
+    else:
+        print('Chiadog requires a positional argument --config')
+        argparse.print_usage()
```
martomi, line 114: Have you tried this: https://stackoverflow.com/a/11155124

martomi: Do we still need that else statement and returning argparse from the function?

martomi (approved):

Issue #77, opened 2021-05-04
Feature Request: Allow event notifications to be customized    


Chardmw:
Currently event notifications are for warning level events. 

Would like to be able to select different warning and info level notifications for each type of notification (Pushover, SMTP, Telegram, etc...)

suckatlife:
Would also be nice to optionally have more than one farm summary notification per day.

PR #132, opened 2021-05-17
Added notification when plot count changes

jamierf:
I added a LOW level notification that fires when plot count increases.  This happens on startup and whenever plot count increases.

### DIFF
--- a/src/chia_log/handlers/condition_checkers/non_decreasing_plots.py
+++ b/src/chia_log/handlers/condition_checkers/non_decreasing_plots.py
@@ -19,11 +19,18 @@ class NonDecreasingPlots(HarvesterConditionChecker):
         self._max_farmed_plots = 0
 
     def check(self, obj: HarvesterActivityMessage) -> Optional[Event]:
+        event = None
         if obj.total_plots_count > self._max_farmed_plots:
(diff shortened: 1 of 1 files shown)


martomi:
Thanks for the proposal, this will unfortunately cause 30+ daily notifications for people plotting 3+ TB per day so I wouldn't like to merge it to the main branch.

Chardmw:
Can the excess notifications be filtered as part of Issue #77.

chadmccune:
I agree, even though I'm plotting about 10TB a day right now, I'd still
like to have this.

jamierf:
I can see handling it either by adding a config flag just for this event to go to notifications or a flag that enables all low level events to go to the notifications.  Perhaps I'll see if I can figure that out later tonight.

andymarden:
I'm using this mod (I added a pull request where I included time since last plot which is nice) - agree that it could be a config flag just for this message defaulted to false for each notification channel.

andymarden (approved):
Also include pull request to add time since last plot

martomi:
We've merged and released this feature with this implementation: https://github.com/martomi/chiadog/pull/307

Now it's possible to configure notification for increases / decreases separately.

PR #133, opened 2021-05-17
Fix harvester activity parser regexp to support more than 9 matching …

ctrlaltdel:
…plots

Without this change, logs like this one were not correctly parsed:

harvester chia.harvester.harvester: INFO     10 plots were eligible for farming c599f64427... Found 0 proofs. Time: 0.45573 s. Total X plots


martomi (approved):
Good catch, thanks! Do you also want to extend that test cases with that? I'll merge this for now as-is in the interest of including it into today's bugfix release.

Issue #116, opened 2021-05-14
Shuffled signage points not correctly handled


martomi:
**Describe the bug**

There are still cases when shuffled signage points during forks are not correctly handled.

The following sequence of events generates 61 skipped SP event.
```
2021-05-13T22:12:25.148 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 60/64: 
2021-05-13T22:12:35.081 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 61/64: 
2021-05-13T22:12:41.911 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 62/64: 
2021-05-13T22:12:49.784 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 63/64: 
2021-05-13T22:12:58.453 full_node chia.full_node.full_node: INFO     ⏲️  Finished sub slot, SP 64/64, 
2021-05-13T22:13:07.659 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 1/64: 
2021-05-13T22:13:14.053 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 2/64: 
2021-05-13T22:13:42.341 full_node chia.full_node.full_node: INFO     ⏲️  Finished sub slot, SP 64/64, 
2021-05-13T22:13:50.884 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 1/64: 
2021-05-13T22:13:58.247 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 2/64: 
2021-05-13T22:14:06.288 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 3/64: 
2021-05-13T22:14:14.181 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 4/64: 
2021-05-13T22:14:22.166 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 5/64: 
2021-05-13T22:14:30.104 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 6/64: 
2021-05-13T22:14:38.398 full_node chia.full_node.full_node: INFO     ⏲️  Finished signage point 7/64: 
```

One way to fix would be comparing expected time for 61 SP (61 times 8-9 seconds) with actual time difference to previous SP.

**Environment:**
 - Chia version: 1.1.5

martomi:
[...]

PR #136, opened 2021-05-17
SignagePoints: Fix wrong handling of fork events

martomi:
* Add test case for network fork
* Fix logic to handle this case properly

Fixes: #116

Should result in less false alarms.


pieterhelsen (approved):
Good to go! 🚀

Issue #146, opened 2021-05-19
Support for pinging a server by IP address?


Foxh0und:
Due to the use of Urllib, we can't open IP addresses for the keep alive. 

`Failed to ping keep-alive: unknown url type: '10.25.160.151' (keep_alive_monitor.py:116)`

**Environment:**
 - OS: Raspbian Buster
 - Python version: 3.7

Foxh0und:
@martomi I have a branch ready to go for this, if I could have permission, that would be great.

martomi:
@Foxh0und You can fork the repository, push your fixes on a branch there and open a PR to this repository directly from the fork. It doesn't require you to have any extra access.

PR #154, opened 2021-05-20
Update: Keep Alive Monitor to use subprocess

Foxh0und:
#146 
Will close out [this issue](https://github.com/martomi/chiadog/issues/146), which will allow us to ping IPv4 addresses.

### DIFF
--- a/src/notifier/keep_alive_monitor.py
+++ b/src/notifier/keep_alive_monitor.py
@@ -1,6 +1,7 @@
 # std
 import logging
-import urllib.request
+import platform
+import subprocess
 from datetime import datetime
 from threading import Thread
 from time import sleep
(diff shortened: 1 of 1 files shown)


src/notifier/keep_alive_monitor.py
```diff
@@ -109,11 +110,14 @@ def _ping_remote(self):
         """
 
         if self._ping_url:
+            param = '-n' if platform.system().lower()=='windows' else '-c'
             logging.debug("Pinging remote keep-alive endpoint")
-            try:
-                urllib.request.urlopen(self._ping_url, timeout=10)
-            except Exception as e:
-                logging.error(f"Failed to ping keep-alive: {e}")
+            command = ['ping', param, '1', self._ping_url]
```
martomi, line 115: What service do you have on the other side checking for those pings?

The publicly available services for health checks usually expose an API GET method that needs to be pinged at an expected interval. I think most people are using this with a service similar to healtchecks.io . And these services aren't compatible with the proposed ping implementation.

Foxh0und: I was just pinging the machine directly, however, now reading, I get your approach. I might set up an api endpoint.

 Thanks.

Foxh0und:
Understanding the use of it now, no need to merge.

Issue #134, opened 2021-05-17
"tail -F" sometime fail to follow new log file


bhy:
On Linux it looks like `tail -F` may not reliably follow new log files. When this happens, chiadog will report "Your harvester appears to be offline", and inspecting the `/proc` file system shows that the `tail` process is still opening the old log file, eg. `.chia/mainnet/log/debug.log.11`. 

**Environment:**
 - OS: Ubuntu 20.04
 - Python version: 3.8.5

pieterhelsen:
Thanks for this insight!
It looks like that's definitive proof that we need to move away from using `tail` and Windows' `Get-Content`.

This seems like it could provide a cross-platform solution for LOCAL harvesters
https://stackoverflow.com/a/43547769 

For remote harvesting, however, this will not work. One possible alternative is Paramiko's `SFTPClient.open` method, but this is slow (as per this [discussion](https://stackoverflow.com/questions/58433996/reading-file-opened-with-python-paramiko-sftpclient-open-method-is-slow)) and probably won't handle log rotations well.

bhy:
Meanwhile I think the following `tail` option might help:

```
       --max-unchanged-stats=N
              with --follow=name, reopen a FILE which has not

              changed size after N (default 5) iterations to see if it has been unlinked or renamed (this  is  the  usual
              case of rotated log files); with inotify, this option is rarely useful
```

pieterhelsen:
It would likely help, but it's still not very pythonic and does not work for cross-platform implementation.
I plan to spend some time on fixing this in the next couple of days.

Gr33nDrag0n69:
This bug is what prevent me to use this software for production monitoring. Overall, very good job. Concerning the issue, I have a couple ideas on how it ccould be at least be dirty fixed before making better code for the long run. The simpliest drity fix for now could be:

When 'Your harvester appears to be offline! No events for the past X seconds.' is detected, the code could just try to reload log consumer first before sending warning.

Gr33nDrag0n69:
Also, on windows, the current never work at all. (As soon as the logrotate code from chia logging module is executed, the harvester offline warning appear and the only is to kill and restart chiadog)

wimsjohn:
I can second this. Using chiadog in WSL for monitoring my windows plotter and farmer. My files split roughly every 22minutes and I am getting spammed with "Your harvester appears to be offline" messages.

pieterhelsen:
There's a branch currently in Pull Request that will likely solve this issue. 
[...]

PR #164, opened 2021-05-21
Improve harvester offline handling

pieterhelsen:
Split up the Windows and Linux/MacOS logic into separate consumers.

Added `pygtail` for Posix and Windows local consumers to enable better log rotation handling. Pygtail does add an artefact file (`debug.log.offset`) to the `.chia/mainnet/log/` folder. This could potentially be moved to the chiadog folder if that's preferable.

Added `retry` to catch Pygtail errors and retry

Added `has_rotated` method for Windows Network consumer which reads log filesize over SSH using Powershell

Log output will look similar to this:
`[2021-05-21 23:12:45] [ WARNING] --- [WinError 2] The system cannot find the file specified: 'C:\\Users\\PieterHelsen\\workspace\\debug.log', retrying in 2 seconds... (api.py:40)`

### DIFF
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -16,7 +16,7 @@ from typing import List, Optional, Tuple
 
 # project
 
-from src.config import check_keys, is_win_platform
+from src.config import check_keys
 from src.util import OS
 
 # lib
@@ -56,6 +56,7 @@ class LogConsumer(ABC):
 class FileLogConsumer(LogConsumer):
     def __init__(self, log_path: Path):
         super().__init__()
+        logging.info("Enabled local file log consumer.")
         self._expanded_log_path = str(log_path.expanduser())
         self._offset_path = PurePath("debug.log.offset")
         self._is_running = True
@@ -72,22 +73,6 @@ class FileLogConsumer(LogConsumer):
         while self._is_running:
             for log_line in Pygtail(self._expanded_log_path, read_from_end=True, offset_file=self._offset_path):
                 self._notify_subscribers(log_line)
-
-
-class PosixFileLogConsumer(FileLogConsumer):
-    """Specific implementation for a simple file consumer for Linux/MacOS"""
-
-    def __init__(self, log_path: Path):
-        logging.info("Enabled Posix file log consumer.")
-        super(PosixFileLogConsumer, self).__init__(log_path)
-
-
-class WindowsFileLogConsumer(FileLogConsumer):
-    """Specific implementation for a simple file consumer for Windows"""
-
-    def __init__(self, log_path: Path):
-        logging.info("Enabled Windows file log consumer.")
-        super(WindowsFileLogConsumer, self).__init__(log_path)
 
 
 class NetworkLogConsumer(LogConsumer):
@@ -228,10 +213,7 @@ def create_log_consumer_from_config(config: dict) -> Optional[LogConsumer]:
         if not check_keys(required_keys=["file_path"], config=enabled_consumer_config):
             return None
 
-        if is_win_platform():
-            return WindowsFileLogConsumer(log_path=Path(enabled_consumer_config["file_path"]))
-        else:
-            return PosixFileLogConsumer(log_path=Path(enabled_consumer_config["file_path"]))
+        return FileLogConsumer(log_path=Path(enabled_consumer_config["file_path"]))
 
     if enabled_consumer == "network_log_consumer":
         if not check_keys(
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -22,7 +22,7 @@ from src.util import OS
 # lib
 import paramiko
 from paramiko.channel import ChannelStdinFile, ChannelStderrFile, ChannelFile
-from pygtail import Pygtail # type: ignore
+from pygtail import Pygtail  # type: ignore
 from retry import retry
 
 
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -57,6 +57,7 @@ class FileLogConsumer(LogConsumer):
     def __init__(self, log_path: Path):
         super().__init__()
         self._expanded_log_path = str(log_path.expanduser())
+        self._offset_path = PurePath('debug.log.offset')
         self._is_running = True
         self._thread = Thread(target=self._consume_loop)
         self._thread.start()
@@ -69,7 +70,7 @@ class FileLogConsumer(LogConsumer):
     @retry((FileNotFoundError, PermissionError), delay=2)
     def _consume_loop(self):
         while self._is_running:
-            for log_line in Pygtail(self._expanded_log_path, read_from_end=True):
+            for log_line in Pygtail(self._expanded_log_path, read_from_end=True, offset_file=self._offset_path):
                 self._notify_subscribers(log_line)
 
 
--- a/src/chia_log/log_consumer.py
+++ b/src/chia_log/log_consumer.py
@@ -57,7 +57,7 @@ class FileLogConsumer(LogConsumer):
     def __init__(self, log_path: Path):
         super().__init__()
         self._expanded_log_path = str(log_path.expanduser())
-        self._offset_path = PurePath('debug.log.offset')
+        self._offset_path = PurePath("debug.log.offset")
         self._is_running = True
         self._thread = Thread(target=self._consume_loop)
         self._thread.start()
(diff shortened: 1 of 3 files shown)


martomi:
@ajacobson FYI this will require giving docker write access to the log folder.

 https://github.com/ajacobson/chiadog-docker/blob/22db43bfee611d31a405b570278ac679dfa9e91a/docker-compose.yaml#L9

src/chia_log/log_consumer.py
```diff
@@ -50,53 +54,69 @@ def _notify_subscribers(self, logs: str):
 
 
 class FileLogConsumer(LogConsumer):
-    """Specific implementation for a simple file consumer"""
-
     def __init__(self, log_path: Path):
-        logging.info("Enabled file log consumer.")
         super().__init__()
         self._log_path = log_path
+        self._expanded_log_path = ""
```
martomi, line 60: Can we initialise this immediately?
```
str(self._log_path.expanduser())
```
It seems then we also don't need to keep `self._log_path` in the state.

src/chia_log/log_consumer.py
```diff
@@ -50,53 +54,69 @@ def _notify_subscribers(self, logs: str):
...
         self._is_running = True
         self._thread = Thread(target=self._consume_loop)
         self._thread.start()
+        self._log_size = 0
 
     def stop(self):
         logging.info("Stopping")
         self._is_running = False
 
+    @retry((FileNotFoundError, PermissionError), delay=2)
```
martomi, line 70: Does this silently retry or it'll print something out? If the user specifies wrong path or has incorrect permissions setup, it should alert them in some way.

Maybe a good idea to set a limit on number of retries, e.g. `tries=3` 🤔

src/chia_log/log_consumer.py
```diff
@@ -50,53 +54,69 @@ def _notify_subscribers(self, logs: str):
...
     def _consume_loop(self):
-        expanded_user_log_path = str(self._log_path.expanduser())
-        logging.info(f"Consuming log file from {expanded_user_log_path}")
+        while self._is_running:
+            for log_line in Pygtail(self._expanded_log_path, read_from_end=True, paranoid=True):
```
martomi, line 73: Do we know if `paranoid=True` is really necessary? I think it might add quite a lot of unnecessary disk writes.

src/chia_log/log_consumer.py
```diff
@@ -50,53 +54,69 @@ def _notify_subscribers(self, logs: str):
...
+
+    def __init__(self, log_path: Path):
+        logging.info("Enabled Posix file log consumer.")
+        super(PosixFileLogConsumer, self).__init__(log_path)
+
+    def _consume_loop(self):
+        self._expanded_log_path = str(self._log_path.expanduser())
+        logging.info(f"Consuming log file from {self._expanded_log_path}")
+
+        super(PosixFileLogConsumer, self)._consume_loop()
```
martomi, line 88: This method implementation seems to be shared between both the Posix and Windows consumers. Let's move it to the parent entirely?

src/chia_log/log_consumer.py
```diff
@@ -109,27 +129,68 @@ def stop(self):
...
+
+class PosixNetworkLogConsumer(NetworkLogConsumer):
+    """Consume logs over SSH from a remote Linux/MacOS harvester"""
+
+    def __init__(self, remote_log_path: PurePath, remote_user: str, remote_host: str, remote_platform: OS):
+        logging.info("Enabled Posix network log consumer.")
+        super(PosixNetworkLogConsumer, self).__init__(remote_log_path, remote_user, remote_host, remote_platform)
+
+    def _consume_loop(self):
+        stdin, stdout, stderr = self._ssh_client.exec_command(f"tail -F {self._remote_log_path}")
```
martomi, line 144:
```
super(WindowsNetworkLogConsumer, self)._consume_loop()
```
seems to be missing

src/chia_log/log_consumer.py
```diff
@@ -109,27 +129,68 @@ def stop(self):
...
+
+    def _consume_loop(self):
+        super(WindowsNetworkLogConsumer, self)._consume_loop()
+
+        stdin, stdout, stderr = self._read_log()
+
+        while self._is_running:
+            if self._has_rotated(self._remote_log_path):
+                sleep(1)
+                self._read_log()
```
martomi, line 166: Think this should be 
```
stdin, stdout, stderr = self._read_log()
```

src/chia_log/log_consumer.py
```diff
@@ -164,32 +225,38 @@ def create_log_consumer_from_config(config: dict) -> Optional[LogConsumer]:
...
     if enabled_consumer == "network_log_consumer":
         if not check_keys(
-            required_keys=["remote_file_path", "remote_host", "remote_user"],
-            config=enabled_consumer_config,
+            required_keys=["remote_file_path", "remote_host", "remote_user"], config=enabled_consumer_config
         ):
             return None
 
-        # default SSH Port : 22
-        remote_port = enabled_consumer_config.get("remote_port", 22)
```
martomi, line 177: Intentionally dropping support for the custom port 🤔

martomi (commented):
Great work 👍 
I'll test the new Posix consumer today!

martomi:
Does this resolve https://github.com/martomi/chiadog/issues/72 🤔

src/chia_log/log_consumer.py
pieterhelsen: This prints out the following in the log
`[2021-05-21 23:12:45] [ WARNING] --- [WinError 2] The system cannot find the file specified: 'C:\\Users\\PieterHelsen\\workspace\\debug.log', retrying in 2 seconds... (api.py:40)`

As per retries, that was how I implemented it first, but since the retry count is on the _consume_loop, this makes it stop working after 3 bad rotations. The retry counter doesn't seem to get reset.

src/chia_log/log_consumer.py
pieterhelsen: I agree, I've tested both and `paranoid` seemed more reliable at picking up errors, but having again retested without paranoid for a bit just now, it seems to work equally reliable, so it might have just been something else that changed while I was busy testing multiple config parameters.

I would suggest to leave it out for now and put it back in if we see any more issues

src/chia_log/log_consumer.py
pieterhelsen: Agree, remnant from my previous implementation

src/chia_log/log_consumer.py
pieterhelsen: Good catch!

src/chia_log/log_consumer.py
pieterhelsen: Wrong cherry pick... reverted

pieterhelsen:
> @ajacobson pinging you here as that might require some changes to Docker. We still need read-only access to the log file itself, but docker should be able to write a file in that same folder (can be temporary, not synced with host machine).
> 
> https://github.com/ajacobson/chiadog-docker/blob/22db43bfee611d31a405b570278ac679dfa9e91a/docker-compose.yaml#L9
> 
> Edit: @pieterhelsen mentioned this file can be moved to chiadog folder. Maybe that'll be preferable for the docker setup?

Added the manual offset to write to within the chiadog root directory. Don't think there's any downsides and might be easier for Docker. 

> Does this resolve #72 🤔

I sure hope so!

pieterhelsen:
This closes #134

src/chia_log/log_consumer.py
```diff
@@ -50,49 +54,56 @@ def _notify_subscribers(self, logs: str):
...
+        logging.info("Enabled Posix file log consumer.")
+        super(PosixFileLogConsumer, self).__init__(log_path)
+
+
+class WindowsFileLogConsumer(FileLogConsumer):
+    """Specific implementation for a simple file consumer for Windows"""
+
+    def __init__(self, log_path: Path):
+        logging.info("Enabled Windows file log consumer.")
+        super(WindowsFileLogConsumer, self).__init__(log_path)
```
martomi, line 90: It just occurred to me that we don't anymore need separate log consumers for PosixFile and WindowsFile anymore?

pieterhelsen: Done! Initially left it in primarily because of consistency with NetworkConsumer, but you're right; it adds unwanted complexity

martomi (approved):
Thanks for addressing all feedback! Looking forward to see if this resolves the issues

PR #168, opened 2021-05-22
Fix black error

pieterhelsen:
When running an unsuccessful `black` format check the script automatically runs the formatter and commits the altered source code.

This currently results in an error:
```
Run black src tests
reformatted src/chia_log/log_consumer.py
All done! ✨ 🍰 ✨
1 file reformatted, 62 files left unchanged.
error: pathspec 'feature/improve-harvester-offline' did not match any file(s) known to git
Error: Process completed with exit code 1.
```

Adding `git fetch` potentially solves this error.


martomi (approved):
Yeah that should do it.

martomi (changes requested):
Actually I think you'll want to run black after checking out the branch. I suggest also breaking the formatting somewhere to test it within this PR.

martomi (approved):
So now I can contribute to the repository while sleeping 😄

PR #174, opened 2021-05-24
NonDecreasingPlots: Increase tolerance threshold to 2

### LINKED ISSUE #58: Number of plots chaning (non_decreasing_plots) to a variable
Hi,

If you are plotting and farming on the same machine the number of plots will go up and down causing an alert.

I have changed my code in non_decreasing_plots.py from:
        if obj.total_plots_count < self._max_farmed_plots:
to
        if abs(obj.total_plots_count - self._max_farmed_plots) >= 5:    

That way if I loose a full drive, or many plots I will get alerted but not be bothered when plot jobs finis.

Reading the value from the config file seems like a better longer term option

Thanks

### LINKED ISSUE #147: Add a "plot count changed" threshold.
I'm using plotman on my plotters and I've enabled the option to sync my plots to a remote machine (my main storage server). Because of this the total plot count on the plotters sometime decrease by 1 (once a plot has been successfully synced). It'd be quite useful to add a config option to specify a threshold when reporting a "plot count changed" warning (e.g. don't emit it if the plot count dropped by only 1 or 2).

What do you think of this?

martomi:
Closes #147
Closes #58

I see little value in making refactors and adding additional
configuration parameters for this. A threshold of 2 should resolve all
issues that are experienced during copying of plots.

Note that 2 is an instantenous threshold since the max value gets reset
after that. Meaning that plot count may drop by 10 without a
notification as long as its dropping in steps of 1.

For transparency of what is happening, I've added an info log that is
always triggered, regardless of the threshold.

The test case is adjusted to contain a scenario with decreasing plot
count by 1 which is expected not to trigger a notification.

### DIFF
--- a/src/chia_log/handlers/condition_checkers/non_decreasing_plots.py
+++ b/src/chia_log/handlers/condition_checkers/non_decreasing_plots.py
@@ -17,6 +17,9 @@ class NonDecreasingPlots(HarvesterConditionChecker):
     def __init__(self):
         logging.info("Enabled check for non-decreasing total plot count.")
         self._max_farmed_plots = 0
+        # When copying plots it's common that plots decrease by 1 temporarily
+        # by setting the default threshold to 2, we can avoid false alarms
+        self._decrease_warn_threshold = 2
 
     def check(self, obj: HarvesterActivityMessage) -> Optional[Event]:
         if obj.total_plots_count > self._max_farmed_plots:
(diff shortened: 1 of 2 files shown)


src/chia_log/handlers/condition_checkers/non_decreasing_plots.py
```diff
@@ -25,14 +28,16 @@ def check(self, obj: HarvesterActivityMessage) -> Optional[Event]:
...
         if obj.total_plots_count < self._max_farmed_plots:
-            message = (
-                f"Disconnected HDD? The total plot count decreased from "
-                f"{self._max_farmed_plots} to {obj.total_plots_count}."
-            )
-            logging.warning(message)
-            event = Event(
-                type=EventType.USER, priority=EventPriority.HIGH, service=EventService.HARVESTER, message=message
-            )
+            logging.info(f"Plots decreased from {self._max_farmed_plots} to {obj.total_plots_count}")
```
pieterhelsen, line 31: Could be an if-else to avoid duplicate information (the same info is repeated, just with slightly different wording and priority)

pieterhelsen (dismissed):
There's an argument to be made for both approaches... I'll leave it up to you to decide. 
For me it's OK to merge either way.

martomi:
Benefit to effort ratio is highest with this approach. The issue has been open for too long, so I want to resolve it and further improvements are welcome by the community as follow-ups!

pieterhelsen (approved):

PR #175, opened 2021-05-24
README: Improvements based on some FAQ

martomi:
- Apply formatting to both WINDOWS.md and README.md
- Add link from README to new Status Reference wiki page
- Add clarifications about multi-harvester setup monitoring
- Add link to disappearing harvester issue on WINDOWS (log rotation)

PR #176, opened 2021-05-24
DailyStats: Add configurable frequency

### LINKED ISSUE #69: Add custom frequency config option for daily stats report & add manual trigger

martomi:
Closes #69

Add option to have daily notifications with higher frequency than every
24 hours. Also add a frequently requested feature to simplify debugging
of the setup by sending immediate and frequent notificaitons.

Will update README for that feature in #175 to simplify conflicts.

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -31,6 +31,7 @@ keep_alive_monitor:
 daily_stats:
   enable: true
   time_of_day: 21
+  frequency_hours: 24
 
 # We support a lot of notifiers, please check the README for more
 # information. You can delete the sections which you aren't using.
--- a/src/chia_log/handlers/daily_stats/stats_manager.py
+++ b/src/chia_log/handlers/daily_stats/stats_manager.py
@@ -26,12 +26,9 @@ class StatsManager:
     """
 
     def __init__(self, config: dict, notify_manager: NotifyManager):
-        try:
-            self._enable = config["enable"]
-            self._time_of_day = config["time_of_day"]
-        except KeyError as key:
-            logging.error(f"Invalid config.yaml. Missing key: {key}")
-            self._enable = False
+        self._enable = config.get("enable", False)
+        self._time_of_day = config.get("time_of_day", 21)
+        self._frequency_hours = config.get("frequency_hours", 24)
 
         if not self._enable:
             logging.warning("Disabled stats and daily notifications")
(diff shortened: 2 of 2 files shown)


martomi:
Also makes #82 obsolete.

src/chia_log/handlers/daily_stats/stats_manager.py
```diff
@@ -96,7 +103,10 @@ def _run_loop(self):
         while self._is_running:
```
pieterhelsen, line 103: L93 reads 'here's what happened in the last 24 hours:'
This is incorrect when the frequency_hours is not set to 24. 
Debug mode should also be handled here.

config-example.yaml
```diff
@@ -31,6 +31,8 @@ keep_alive_monitor:
 daily_stats:
   enable: true
   time_of_day: 21
+  frequency_hours: 24
+  debug_mode: false # Send notifications every 5 minutes
```
pieterhelsen, line 35: I'm not a fan of this implementation, since in 99% of use cases you will only run this once and then have to change the config and restart chiadog. Furthermore, it feels wrong to add a debug mode specifically for the `daily_stats` config. 

I would prefer to add this as an optional argument to the argument parser that sends a one-time notification instead. There's also no use sending the daily_stats specifically, as they will only contain 5 minutes worth of data.

pieterhelsen (changes requested):
Good implementation of the hourly frequency if you are able to change the 'here's what happened in the last X hours:' line
I would suggest to leave out debug mode for now and take the argparser approach instead.

src/chia_log/handlers/daily_stats/stats_manager.py
martomi: Good catch

config-example.yaml
martomi: Agree, will remove.

pieterhelsen (approved):
🥳

PR #179, opened 2021-05-24
LastFarmEvent: Increase warn threshold and improve message

martomi:
Since the network experiences delays up to 70-80 seconds occasionally,
increase warning threshold to 90 to prevent false alarms.

Also change text of info text to avoid frequent questions and concerns.

### DIFF
--- a/src/chia_log/handlers/condition_checkers/time_since_last_farm_event.py
+++ b/src/chia_log/handlers/condition_checkers/time_since_last_farm_event.py
@@ -21,7 +21,7 @@ class TimeSinceLastFarmEvent(HarvesterConditionChecker):
     def __init__(self):
         logging.info("Enabled check for farming events.")
         self._info_threshold = 30
-        self._warning_threshold = 60
+        self._warning_threshold = 90
         self._last_timestamp = None
 
     def check(self, obj: HarvesterActivityMessage) -> Optional[Event]:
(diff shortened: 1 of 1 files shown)


src/chia_log/handlers/condition_checkers/time_since_last_farm_event.py
```diff
@@ -44,7 +44,11 @@ def check(self, obj: HarvesterActivityMessage) -> Optional[Event]:
         elif seconds_since_last > self._info_threshold:
             # This threshold seems to be surpassed multiple times per day
             # on the current network. So it only generates an INFO log.
-            logging.info(f"Last farming event was {seconds_since_last} seconds ago. This is unusual.")
+            logging.info(
+                f"Last farming event was {seconds_since_last} seconds ago."
+                " Usually every 9-10 seconds. No reason to worry if it happens "
```
pieterhelsen, line 49: I would prefer to add this to the wiki and keep the info here clean and concise.

pieterhelsen (changes requested):

src/chia_log/handlers/condition_checkers/time_since_last_farm_event.py
martomi: I'd agree for notification texts, but this is part of the actual process logs so we can afford more verbosity, I think. 🤔 I'd like that at least to some degree users can tell directly from the log message what they're seeing, not always having to refer to the wiki page.

martomi: Either we do it this way or we hide this log on DEBUG level. Because many users will see this and won't see the wiki page.

pieterhelsen (approved):

PR #183, opened 2021-05-24
Events: Bring back the found proof notifications

### LINKED ISSUE #167: Support reward notifications for more scenarios
I believe that if you follow the best practice of using a separate (offline) wallet (describe here https://github.com/Chia-Network/chia-blockchain/wiki/Good-Security-Practices-on-Many-Machines#keep-your-wallet-separate) you won't get notified (because the wallet logs won't contain the lines for the correct wallet)? Maybe it's a good idea to have a separate notification for when you have won and another one that watches the wallet logs?

### LINKED ISSUE #182: Found a proof, no email
Latest versions of everything (linux, PIP, chia and chiadog). It is a remote harvester monitoring a local (local to the harvester) farm. I noticed a new block reward in my chia GUI but had not received a chiadog email notification. I logged into the remote harvester and see the "Found a proof!" message in the chiadog window. But no email came through. I do receive the daily stats email. I checked the config file and under notifier -> email, both daily_stats and wallet_events are set to true. Is there a place to look for error messages? Maybe a log file somewhere? I couldn't find one.

martomi:
closes #167
closes #182

The wallet events that were supposed to replace these notifications do
not work in many of the scenarious that chiadog operates in.

This commit reverses the removal of the found proof notifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ helps with automated monitoring and sends you a mobile notification in case some
 | Harvester | Experiencing networking issues? Harvester did not participate in any challenge for 120 seconds. It's now working again. | NORMAL |
 | Harvester | Seeking plots took too long: 21.42 seconds! | NORMAL |
 | Full Node | Experiencing networking issues? Skipped 42 signage points! | NORMAL |
+| Harvester | Found 1 proof(s)! | LOW |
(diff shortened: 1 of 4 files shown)


martomi:
If we merge this, we'll need to add to the wiki that this not guarantee a farmed block + reward.

pieterhelsen (approved):
Good to go!

PR #184, opened 2021-05-24
Add Link to wiki Status Reference

### LINKED ISSUE #159: [Feature Request] Add detailed error description and potential fixes
It would be good to have a detailed wiki page that, for each possible error message from Chiadog, lists:

- The priority of the issue (should I worry, or not)
- The reasons why this issue might occur
- Potential fixes to this issue

Once this is in place, we can link to the specific page from the Chiadog notifications

pieterhelsen:
Updated Telegram notifier to not generate a URL preview
Tested on Slack, Discord and Telegram. 

Closes #159


martomi:
What do you think about adding a function for suffix, similar to title here https://github.com/martomi/chiadog/blob/main/src/notifier/__init__.py#L78

And then appending that suffix to all user events? Then it'll be a single place to change it and we don't need to remember to append it to any new events added in future.

PR #185, opened 2021-05-25
Added startup switch that trigger daily_stats after 30 seconds

pieterhelsen:


src/chia_log/handlers/daily_stats/stats_manager.py
```diff
@@ -82,11 +90,16 @@ def consume_signage_point_messages(self, objects: List[FinishedSignagePointMessa
                 for obj in objects:
                     stat_acc.consume(obj)
 
-    def _send_daily_notification(self):
-        summary = f"Hello farmer! 👋 Here's what happened in the last {self._frequency_hours} hours:\n"
+    def _send_daily_notification(self, startup: bool = False):
```
pieterhelsen, line 93: Another approach would be to copy this code to split this code into `_send_daily_notification()` and `_send_test_notification()`. Not sure yet which I prefer.

main.py
```diff
@@ -23,7 +23,8 @@ def parse_arguments() -> Tuple[ArgumentParser, Namespace]:
     )
     group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument('--config', type=str, help="path to config.yaml")
-    group.add_argument('--version', action='store_true')
+    group.add_argument('--version', action='store_true', help="show current version of chiadog")
+    parser.add_argument('--startup', action='store_true', help="trigger startup notification after 1 minute")
```
martomi, line 27: Should be 30 seconds

src/chia_log/handlers/daily_stats/stats_manager.py
```diff
@@ -49,10 +50,17 @@ def __init__(self, config: dict, notify_manager: NotifyManager):
...
         )
+
+        if self._startup:
+            logging.info("An initial startup notification will be sent in 30 seconds")
+
         self._datetime_next_summary = datetime.now().replace(hour=self._time_of_day, minute=0, second=0, microsecond=0)
         while datetime.now() > self._datetime_next_summary:
             self._datetime_next_summary += timedelta(hours=self._frequency_hours)
 
+        self._datetime_startup_summary = datetime.now().replace(microsecond=0)
```
martomi, line 61: `replace(microsecond=0)` this is not strictly necessary but ok to leave 😄

src/chia_log/handlers/daily_stats/stats_manager.py
```diff
@@ -97,6 +110,9 @@ def _run_loop(self):
             if datetime.now() > self._datetime_next_summary:
                 self._send_daily_notification()
                 self._datetime_next_summary += timedelta(hours=self._frequency_hours)
+            if self._startup and datetime.now() > self._datetime_startup_summary:
+                self._send_daily_notification(startup=True)
+                self._startup = False
```
martomi, line 115: See no need for an if-statement here, the flag can be reset directly in `_send_daily_notification`

src/chia_log/handlers/daily_stats/stats_manager.py
martomi: I would probably have left it without if-else statements for simplicity. Not resetting the stats only adds another 30 seconds of data for the next daily notification which isn't substantial to sweat over and the summary text change maybe nice but it'll be obvious either way I think.

Anyway if you decide to keep it with the extra handling, I suggest to directly read the `self._startup` variable without passing as parameter and reset it here as well, so that all handling is in the same function and not also spread over in the `_run_loop`.

martomi (changes requested):

PR #189, opened 2021-05-25
Dev Branch - main.py logging issue

ninthwalker:
Wasn't sure where to mention this, so just doing a PR with this note since I don't know a lot about python.
In the new DEV branch, on line 54 in main.py you added this line:

`logging.info(f"Starting Chiadog ({version()})")`

I did not test in Linux, but at least in windows it will not start with that line and throws an error about 'file not found'.
Commenting out that line allowed the Dev branch to work. Just thought I would mention it.


martomi:
@ninthwalker Thanks for the report. I suppose you have downloaded `chiadog` without git? That's one failure scenario that I can imagine where this wouldn't work.

P.S. The right place would have been under [Issues](https://github.com/martomi/chiadog/issues)

martomi:
Linked PR should resolve this.

PR #193, opened 2021-05-26
Fix install script package checks

bradclow:
Removes the return statement that always causes the packages
check to fail.

Recognises removed packages should be installed again.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,8 @@
+#!/bin/bash
+
 # Function that checks if a package is installed locally
 function package_installed() {
-  return dpkg -l "$1" &> /dev/null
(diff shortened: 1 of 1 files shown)


bradclow:
The current install.sh script always does the apt-get update and install, even if the packages are already installed.
It is caused by the incorrect return statement in function package_installed. The redirect to /dev/null masks the error.

Also found that dpkg -l had a successful exit code even after a package has been removed. dpkg -s handled the situation better.

Tested on Ubuntu 20.04.2 LTS.

Also added the hashbang at the top for good form.

martomi (approved):
Thanks! 👍

PR #195, opened 2021-05-26
Version: Fix zombie process resulting from Popen

### LINKED ISSUE #194: Defunct zombie git process
Hi, I just noticed a git zombie process started by chiadog.

This is from my processlist in Ubuntu
userx     19328    3325  0 14:47 pts/1    00:00:00 /bin/bash
userx     19329   19328  1 14:47 pts/1    00:00:00 python3 main.py --config config.yaml
userx     19330   19329  0 14:47 pts/1    00:00:00 [git] **\<defunct>**

Having a look into main.py tells me the reason should be this part in version():
```
command_args = ["git", "describe", "--tags"]
f = subprocess.Popen(command_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
return f.stdout.readline().decode(encoding="utf-8").rstrip()
```
I guess the git subprocess won't die since the stderr is never consumed.

See here:
https://www.reddit.com/r/learnpython/comments/3lzv4r/subprocess_becomes_zombie_why/
https://stackoverflow.com/questions/2760652/how-to-kill-or-avoid-zombie-processes-with-subprocess-module/2761237#2761237
"Not using Popen.communicate() or call() will result in a zombie process."

It's not a big deal since this only happens once at start, but I'd prefer to not have any permanent zombie processes.
Finally, this should be a very easy and quick fix!

pieterhelsen:
See [Stackoverflow](https://stackoverflow.com/questions/2760652/how-to-kill-or-avoid-zombie-processes-with-subprocess-module/2761237#2761237) for detailed solution

I was unable to reproduce this issue on my machines (both Windows and Linux), so feedback is welcome!

This closes #194

### DIFF
--- a/main.py
+++ b/main.py
@@ -97,7 +97,8 @@ def version():
     try:
         command_args = ["git", "describe", "--tags"]
         f = subprocess.Popen(command_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        return f.stdout.readline().decode(encoding="utf-8").rstrip()
+        stdout, stderr = f.communicate()
+        return stdout.decode(encoding="utf-8").rstrip()
     except:
         return "unknown"
 


main.py
```diff
@@ -97,7 +97,8 @@ def version():
     try:
         command_args = ["git", "describe", "--tags"]
         f = subprocess.Popen(command_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        return f.stdout.readline().decode(encoding="utf-8").rstrip()
+        stdout, stderr = f.communicate()
```
martomi, line 100: There's [subprocess.check_output](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) function which should be exactly what we need here. (it does Popen and communicate calls for you)

djdookie (approved):
Confirming that this fixes the issue.

martomi (approved):

PR #214, opened 2021-05-30
Bugfix: Remove corrupt pygtail offset

### LINKED ISSUE #205: Startup fails after hard reboot of computer
Startup fails after hard reboot of computer.  

When the computer does a hard reset the debug.log.offset file is not closed and removed. The startup of CHIADOG then hangs. See keybase://chat/chia_network.public#chiadog/2744 for example of the error received and the current fix action. 

Can there be a check during startup for this file and remove it if it exists before it hangs? 

**Environment:**
 - OS: Windows 10
 - Python version:
 - PIP version:
 - Chia version: 1.1.6

pieterhelsen:
Closes #205

### DIFF
--- a/main.py
+++ b/main.py
@@ -53,6 +53,10 @@ def init(config:Config):
 
     logging.info(f"Starting Chiadog ({version()})")
 
+    # Remove the Pygtail offset file if it exists
+    # Fixes a bug where hard reboot corrupts the offset file
+    Config.get_offset().unlink(missing_ok=True)
+
     # Create log consumer based on provided configuration
     chia_logs_config = config.get_chia_logs_config()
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -53,6 +53,10 @@ def init(config:Config):
 
     logging.info(f"Starting Chiadog ({version()})")
 
+    # Remove the Pygtail offset file if it exists
+    # Fixes a bug where hard reboot corrupts the offset file
+    Config.get_offset().unlink(missing_ok=True)
```
martomi, line 58: Nitpick: I'd rename the `get_offset` function to `get_log_offset_path`. It'd be a bit more descriptive and clear from the function call what happens here.

martomi (approved):
Thanks!

PR #216, opened 2021-06-01
fix _has_rotated windows powershell.exe

mezzovide:
add missing ")"


pieterhelsen (changes requested):
Thanks for spotting this bug! The first bracket is actually extraneous and can be removed. 
`f"powershell.exe Write-Host(Get-Item {str(path)}).length"`

pieterhelsen (approved):

martomi (approved):
Thanks!

jiimbo420:
Thanks for fix

PR #237, opened 2021-06-13
allow specify hh:mm for daily stats

kanasite:
Allow user to specify the HH:MM for the daily stats (rather than waiting for next hour)

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -30,8 +30,7 @@ keep_alive_monitor:
 # on your farm performance at the specified time of the day.
 daily_stats:
   enable: true
-  notify_hour: 21
-  notify_minute: 0
+  time_of_day: "21:00"
   frequency_hours: 24
 
 # We support a lot of notifiers, please check the README for more
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -30,7 +30,8 @@ keep_alive_monitor:
 # on your farm performance at the specified time of the day.
 daily_stats:
   enable: true
-  time_of_day: 21
+  notify_hour: 21
+  notify_minute: 0
   frequency_hours: 24
 
 # We support a lot of notifiers, please check the README for more
--- a/src/chia_log/handlers/daily_stats/stats_manager.py
+++ b/src/chia_log/handlers/daily_stats/stats_manager.py
@@ -2,7 +2,7 @@
 import logging
 import re
 from datetime import datetime, timedelta
-from typing import List
+from typing import List, Union
 from threading import Thread
 from time import sleep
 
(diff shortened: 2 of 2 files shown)


Zenedith (approved):

config-example.yaml
```diff
@@ -30,7 +30,8 @@ keep_alive_monitor:
 # on your farm performance at the specified time of the day.
 daily_stats:
   enable: true
-  time_of_day: 21
+  notify_hour: 21
+  notify_minute: 0
```
martomi, line 34: Thanks for the proposed change.

I think this won't justify breaking backwards compatibility for all existing configs however. (We don't have auto-config migration). I'd be more willing to merge if the implementation supports old style + maybe new style with parsing of the minutes from the time_of_day string which might either be `21` or `21:00`. 

What do you think?

martomi: Also I suspect the main use-case here is to quickly check notifications are running. We've previously discussed it will be a good idea to have a config flag that will send a daily notification 1 minute after startup: https://github.com/martomi/chiadog/pull/176#discussion_r638080060

pieterhelsen: There's a branch in the works for this, but I haven't had time to properly finish it yet:
https://github.com/martomi/chiadog/tree/feature/startup-switch

kanasite: @martomi thanks for the feedback, I made a quick change so that it doesn't break the backward compatibility of the configs. For me, it's more of providing the flexibility in setting the time :)

config-example.yaml
```diff
@@ -30,7 +30,7 @@ keep_alive_monitor:
 # on your farm performance at the specified time of the day.
 daily_stats:
   enable: true
-  time_of_day: 21
+  time_of_day: "21:00"
```
kanasite, line 33: The existing integer value still works, hence may not necessary changing this to 21:00, but I find it more intuitive defining hh:mm

src/chia_log/handlers/daily_stats/stats_manager.py
```diff
@@ -101,3 +103,20 @@ def _run_loop(self):
...
+            if re.match('(?:[01]\d|2[0123]):(?:[012345]\d)', value) is not None:
+                if hm == "hour":
+                    return int(value[:2])
+                elif hm == "minute":
+                    return int(value[-2:])
+            else:
+                if hm == "hour":
+                    return 21
+                elif hm == "minute":
+                    return 0
```
martomi, line 122: This method is a bit difficult to read and maintain for its functionality. (too many if-statements)
I'd propose to simplify it by returning a dict or tuple with hour & minutes instead.

It's also better design to explicitly specify the default value as an optional argument so it's not hidden away in the implementation and can be more easily changed. Please also add type annotations for function arguments.

I propose roughly something like:

```
from typing import Union

...

def _parse_notify_time(self, value: Union[str, int], default: dict = {'hour': 21, 'minute': 0}) -> dict:
    if type(value) == int:
        return {'hour': value, 'minute': 0}

    match = re.match('(?:[01]\d|2[0-3]):(?:[0-5]\d)', value)
    if match:
        return {'hour': int(value[:2]), 'minute': int(value[-2:])}

    return default
```

martomi (changes requested):
Thanks for implementing the feedback, I like the approach now.

Got some more feedback on the actual implementation :-)

src/chia_log/handlers/daily_stats/stats_manager.py
kanasite: @martomi Sorry about that, my first time into python. I've refactored it and thanks for the feedback!

martomi (approved):

martomi (approved):

PR #244, opened 2021-06-22
Integration: Grafana Annotations API

pieterhelsen:
This creates annotations on Grafana dashboards whenever Chiadog triggers an alert. 

One problem I see currently with this implementation is that it triggers this error:
`[2021-06-22 23:39:14] [   ERROR] --- Invalid config.yaml. Missing key: 'daily_stats' (__init__.py:76)`

This is because daily stats (and wallet events) should not trigger an annotation to be added to a Grafana dashboard (you typically only want to trigger an annotation when something is wrong), so I have left them out of the config-example. 

An example of what an annotation looks like:
image

### DIFF
--- a/tests/notifier/test_grafana_notifier.py
+++ b/tests/notifier/test_grafana_notifier.py
@@ -1,0 +1,46 @@
+# std
+import os
+import unittest
+
+# project
+from src.notifier.grafana_notifier import GrafanaNotifier
+from .dummy_events import DummyEvents
+
+
+class TestGrafanaNotifier(unittest.TestCase):
+    def setUp(self) -> None:
+        base_url = os.getenv("GRAFANA_BASE_URL")
+        api_token = os.getenv("GRAFANA_API_TOKEN")
+        self.assertIsNotNone(base_url, "You must export GRAFANA_BASE_URL as env variable")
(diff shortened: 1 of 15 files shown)


tests/notifier/test_grafana_notifier.py
```diff
@@ -0,0 +1,46 @@
...
+                    "base_url": base_url,
+                    "api_token": api_token,
+                },
+            },
+        )
+
+    @unittest.skipUnless(
+        os.getenv("GRAFANA_BASE_URL") and os.getenv("GRAFANA_BASE_URL"), "Run only if credentials available"
+    )
+    def testSlackLowPriorityNotifications(self):
```
martomi, line 30: method name still refers to slack 😄

martomi (changes requested):
There are two fairly easy ways to resolve the ERROR about the missing key.

Without changes to codebase you can add to the config the `daily_stats` and `wallet_events` and set them to `false` by default. Then they won't appear on Grafana and also won't trigger error.

Alternatively, change the logic here to make the keys optional and default to False. E.g. `config.get("daily_stats", False)`. 

https://github.com/martomi/chiadog/blob/429b90cf2d3f74a885d4ffcb748fb63b1521e48b/src/notifier/__init__.py#L70

martomi (approved):
📈

PR #251, opened 2021-07-05
Fix empty log for running in background mode

maxbanton:


pieterhelsen (approved):
Hi Max,

Thanks for your pull request! This seems to check out :) 
Couple of things to take into account should you submit more PRs in the future. 
As per the [contribution guidelines](https://github.com/martomi/chiadog/blob/main/CONTRIBUTING.md#how-to-contribute), make sure to:

- Create a descriptive bugfix or feature branch (e.g. in this case instead of `patch-1` you would use `bugfix/documentation-nohup-empty-log` or something similar) 
- Point your pull request at the `dev` branch, instead of the `main` branch

martomi:
Hi @maxbanton,

this looks good to merge, I think capturing stderr is a good idea in general. What I don't understand is the wording of the commit message. Can you describe what is the actual fix here? All relevant output goes on stdout currently so this is not fixing empty log as the commit message claims.

maxbanton:
Hi @martomi. Sorry for late reply and messy commit message, proposed change fixes empty log when running in background mode using instructions from Readme. I mean when I added 2>&1, log entries started to appear in the log file, till that it was empty. Probably it's because I'm running chiadog from cron job on reboot event. As for now my crontab entry looks like this
`@reboot . /home/username/chiadog/venv/bin/activate && nohup python3 -u /home/username/chiadog/main.py --config /home/username/chiadog/config.yaml > /home/username/chiadog/output.log 2>&1 &` and it starts automatically every time after reboot.

PR #254, opened 2021-07-08
Update Signage Point regex to comply with Windows log format

### LINKED ISSUE #253: [BUG] Signage Points are not parsed on Windows
While running CHIA in windows, chiadog seems to miss INFO about signage points. I think the issue might be related to a missing symbol in windows chia debug log file related to signage points vs. in linux. 

Chiadog SPs

pieterhelsen:
Fixes #253 

See #253 for details.


martomi (approved):
Very good catch, thanks Pieter! 🎉

PR #265, opened 2021-07-15
Update log offset unlink for compatibility with Python 3.7

### LINKED ISSUE #264: TypeError: unlink() got an unexpected keyword argument 'missing_ok'
**Describe the bug**
In main.py, line 58:
 Config.get_log_offset_path().unlink(missing_ok=True)

Python 3.7.3 hasn't implemented this parameter, cf: https://stackoverflow.com/questions/42636018/python-difference-between-os-remove-and-os-unlink-and-which-one-to-use

"In Python v3.8, a missing_ok argument was added to the Path.unlink() function. When *missing_ok* == True, an exception will not be raised if the file doesn't exist before trying to remove it."

Debian still uses python 3.7.3, I hope they would upgrade soon :)

A clear and concise description of what the bug is and how it can be reproduced.

**Environment:**
 - OS: Linux Debian 10 (Raspbian)
 - Python version: 3.7.3
 - PIP version:
 - Chia version: 1.2.2
 - Chiadog version:  0.6.1
 - Harvester: remote OR local

pieterhelsen:
missing_ok was added in Python 3.8.
Replaced with path.exists() check prior to removal

https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink

Closes #264

### DIFF
--- a/main.py
+++ b/main.py
@@ -55,7 +55,9 @@ def init(config:Config):
 
     # Remove the Pygtail offset file if it exists
     # Fixes a bug where hard reboot corrupts the offset file
-    Config.get_log_offset_path().unlink(missing_ok=True)
+    offset = Config.get_log_offset_path()
+    if offset.exists():
+        offset.unlink()
 
     # Create log consumer based on provided configuration
     chia_logs_config = config.get_chia_logs_config()


martomi (approved):
Thanks!

PR #267, opened 2021-07-15
Add type dependencies for mypy

pieterhelsen:


CONTRIBUTING.md
```diff
@@ -35,6 +35,7 @@ Before submitting a PR make sure that your feature is covered with tests.
 
 ```
 pip3 install black flake8 mypy
+pip3 install types-PyYAML types-python-dateutil types-paramiko types-retry
```
martomi, line 38: Do you know why it suddenly became necessary to install these separately?

martomi (approved):
Thanks!

CONTRIBUTING.md
pieterhelsen: Think it's related to this blogpost:
https://mypy-lang.blogspot.com/2021/05/the-upcoming-switch-to-modular-typeshed.html

martomi: Seems there's also the command `mypy --install-types` which automatically detects and installs the missing types. Something for the future perhaps.

PR #278, opened 2021-08-11
add block found tests

kanasite:
simple test case for block found event


tests/chia_log/handlers/test_block_found_hander.py
```diff
@@ -0,0 +1,30 @@
+# std
```
martomi, line 1: Typo in filename `hander` -> `handler`

martomi (approved):
Thanks for following up with test cases @kanasite, looks good!
Would be great if you can only fix the typo in filename :-)

tests/chia_log/handlers/test_block_found_hander.py
kanasite: oops!

martomi (approved):

PR #293, opened 2021-09-18
fix blockfound parser

kanasite:
@martomi I farmed a pool block today realized the parser is not recognizing, I've updated the regex at least for ubuntu's log

### DIFF
--- a/src/chia_log/parsers/block_parser.py
+++ b/src/chia_log/parsers/block_parser.py
@@ -27,7 +27,7 @@ class BlockParser:
     def __init__(self):
         logging.info("Enabled parser for block found stats.")
         self._regex = re.compile(
-            r"([0-9:.]*) full_node (?:src|chia).full_node.full_node\s*: INFO\s* ((?:🍀|.)\s* Farmed unfinished_block)"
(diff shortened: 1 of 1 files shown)


src/chia_log/parsers/block_parser.py
```diff
@@ -27,7 +27,7 @@ class BlockParser:
     def __init__(self):
         logging.info("Enabled parser for block found stats.")
         self._regex = re.compile(
-            r"([0-9:.]*) full_node (?:src|chia).full_node.full_node\s*: INFO\s* ((?:🍀|.)\s* Farmed unfinished_block)"
+            r"([0-9:.]*) full_node (?:src|chia).full_node.full_node\s*: INFO\s* ((?:🍀 ️|.)\s*Farmed unfinished_block)"
```
martomi, line 30: What's this character after the 🍀? Looks like empty space but it's not. Otherwise `\s*` will cover it.

It'd be good if anyone else can confirm if they have that in their output before we merge this. Also adding an example log and test-case for that parser will prevent similar issues in future. Similar to [already existing log tests](https://github.com/martomi/chiadog/tree/main/tests/chia_log/logs)

martomi: Seems to be this character https://www.fileformat.info/info/unicode/char/fe0f/index.htm

martomi (approved):

PR #314, opened 2021-11-06
Add config option to disable SMTP auth for basic postfix setup

unsocial-bleach:
Sometimes it is useful to disable SMTP authentication for sending emails. For example, if someone is using a local `postfix` server, the default install doesn't require/enable auth. 

When this new config option is set to `enable_smtp_auth: false`, the `username_smtp` and `password_smtp` config options are ignored.


martomi (approved):
Looks good!

PR #319, opened 2021-12-12
Sandboxed systemd service

### LINKED ISSUE #312: Add a systemd service for running this in the background easily on Debian/Ubuntu systems
Please add a chiadog.service systemd service file.

alexking:
Should resolve #312, relates to discussion #51. Attempts to sandbox chiadog as a systemd service, and safeguard any sensitive chia files in `.chia/mainnet` that we don't need access to. 

* Add a scripts/linux/chiadog.service systemd example that attempts to run chiadog in a more isolated environment. Create a new limited user each run (making much of the filesystem readonly), and set `.chia/mainnet` folders other than `log` to inaccessible.
* Move the offset file - previously, the `debug.log.offset` file was kept in the chiadog directory, but when running in a read only filesystem we can't write there. Create a temporary directory when running, and store the offset file there. This also means we no longer need to delete the offset file on startup.

A good way to test this is to switch out the `ExecStart` for an `ls -lh .chia/mainnet` and see what files are accessible inside the environment, or try to run `venv/bin/chia wallet show` inside and make sure it fails. Ideally I'd like to have chiadog show a warning when it starts up if it's able to access configuration directories, but I figure that could be added in another PR. 

There are a bunch of different strategies in systemd for setting this up, I went with what seems to be the most compatible version (compatible with systemd 235, which I believe is available in Ubuntu 18), explicitly declaring `InaccessiblePaths`. The other way, using `TemporarilyFilesystem` to block everything out and binding the one file we need, is maybe a bit more elegant, but needs at least systemd 238 (around Ubuntu 20). Since it's not possible to say "make everything except the logs folder inaccessible" however (declaring `.chia` as inaccessible makes all child directories inaccessible and can't be overridden by other directives), we need to remember to add any new paths if new versions of chia add additional directories we don't want access to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,16 +196,19 @@ nohup python3 -u main.py --config config.yaml > output.log 2>&1 &
 To stop chiadog, you can find the Process ID (PID) via `ps aux | grep main.py` and then softly interrupt the process
 with `kill -SIGINT <pid_here>`.
 
+You can also run chiadog as a sandboxed systemd service.
+
 ## Running `chiadog` in sandboxed environment
 
-We're in the early stages of exploring the best way to provide easy to setup sandboxed environment where the `chiadog`
-process is completely isolated from potentially accessing your private keys. Contributions in that direction are very
-welcome. Meanwhile you can check out @ajacobson repository
-for [chiadog-docker](https://github.com/ajacobson/chiadog-docker).
+We're still exploring the best way to provide easy to setup sandboxed environment where the `chiadog` process is
+completely isolated from potentially accessing your private keys. Contributions in that direction are very welcome.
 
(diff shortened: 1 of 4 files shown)


martomi:
Hi @alexking, 

this was really delightful to review, really appreciated the detailed documentation and careful considerations. 👍 💯 

I'll go ahead and merge this into `dev`, and it'll make it into the final release for 2021 :-)

Thanks!

martomi (approved):

alexking:
Thanks! Also I just noticed that Chia [is migrating their keys to](https://github.com/Chia-Network/chia-blockchain/wiki/Passphrase-Protected-Chia-Keys-and-Key-Storage-Migration) `~/.chia_keys`, so I'll submit another to include that as inaccessible.

PR #320, opened 2021-12-18
Add ~/.chia_keys to list of InaccessiblePaths in systemd

alexking:
Update for #319 - Chia is [migrating their keyfiles](https://github.com/Chia-Network/chia-blockchain/wiki/Passphrase-Protected-Chia-Keys-and-Key-Storage-Migration) to `~/.chia_keys` (I was a version behind so I hadn't noticed - it appears existing keys will remain in `~/.chia/config`, but with fresh installs or new keys, they'll move to `chia_keys`). Block this folder from access in the systemd service.


martomi (approved):
Thanks for noticing that!

PR #323, opened 2022-01-09
Inhouse docker image builds

Artanicus:
This hugely builds on the work by @ajacobson (https://github.com/ajacobson/chiadog-docker) and no disrespect is intended.

Proposing merge to `main` instead of `dev` since this does not in any way modify how Chiadog works and will then enable images to be built for any and all branches regardless of when dev is merged to main.

## Major benefits of inhousing:
  - Can easily build any state of the repo regardless of branch or
      mid-development into a functional image without mucking about with an
      external repo. Doing other chiadog development made me frustrated enough to also fix this. :-)
  - Faster builds when the source code is already present without an extra clone
  - Automated Docker image builds into the free GitHub image repository. Official images triggered by official tags & releases, but also dev images any any other branch provide repeatable and traceable pinning of images.
    
## Minor fixes added as well:
  - entrypoint.sh was missing a shebang
   - Simplified entrypoint running now that the script has a proper shebang
   - Simplified logic by switching WORKDIR earlier
   - Add a .dockerignore file to prune dirty state and useless weight from
      the container.
    
## Does not include:
  - Any documentation. Docker is an advanced option in any case but this
      should be added as a followup.

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -1,0 +1,55 @@
+name: Docker image
+
+on: push
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Get the release channel
+        id: get_channel
+        shell: bash
+        run: |
+          if [[ "$GITHUB_REF" == 'refs/heads/main' ]]; then
+            echo ::set-output name=channel::"latest"
+            echo ::set-output name=version::${GITHUB_REF/refs\/heads\//}-${GITHUB_SHA::6}
+          elif [[ "$GITHUB_REF" == "refs/heads/"* ]]; then
+            echo ::set-output name=version::${GITHUB_REF/refs\/heads\//}-${GITHUB_SHA::6}
+          elif [[ "$GITHUB_REF" == "refs/tags/"* ]]; then
+            echo ::set-output name=channel::${GITHUB_REF/refs\/tags\//}
+          fi
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ steps.get_channel.outputs.channel }}
+            type=raw,value=${{ steps.get_channel.outputs.version }}
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,6 @@
 # secrets
-*.yaml
-*.yml
-!config-example.yaml
-!.github/**/*.yaml
-!.github/**/*.yml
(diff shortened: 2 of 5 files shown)


Artanicus:
Oops, let me rebase this from dev -> main.

Artanicus:
Example image has been automatically built in my fork: https://github.com/Artanicus/chiadog/pkgs/container/chiadog .. this shows the logic for non-special branches with the image name becoming:
`ghcr.io/artanicus/chiadog:inhouse_docker-26c7b8`

.gitignore
```diff
@@ -1,6 +1,9 @@
 # secrets
-config.yaml
-config.*.yaml
+*.yaml
+*.yml
+!config-example.yaml
+!.github/**/*.yaml
+!.github/**/*.yml
```
martomi, line 6: I would actually prefer to not ignore all `*.yaml` files. They are generally useful to have in the repository, and  as we see `.github` also has them. The use-case for having randomly named `config.yaml` files is too exotic to justify general ignore rule. The original version already supports `config.dev.yaml`, `config.prod.yaml` etc.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ENV TZ=UTC
+
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y python3 bash ca-certificates git openssl python3-pip build-essential python3-dev python3.8-venv python3.8-distutils python-is-python3 tzdata
+
+RUN dpkg-reconfigure -f noninteractive tzdata
+
+RUN apt-get install -y libsodium-dev
+RUN SODIUM_INSTALL=system pip3 install pynacl
```
martomi, line 13: Why do we install `libsodium` separately from `pynacl`?

From [docs](https://pynacl.readthedocs.io/en/latest/install/) it seems like `pip install pynacl` should replace both lines above.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu:latest
+
+ENV config_dir=/root/.chiadog/config.yaml
+ENV TZ=UTC
```
martomi, line 4: nitpick: inconsistent letter casing for ENV variables.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
martomi, line 6: Is this necessary considering this is also executed inside `entrypoint.sh` ?

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,55 @@
+name: Docker image
+
+on: push
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
```
martomi, line 10: I like that automated workflow 👍

martomi (commented):
Thanks for the contribution @Artanicus !

I agree with the benefits you listed and particularly like the automated build workflow 👍 

Please check some feedback I have below.

.gitignore
Artanicus: Fair enough, will revert this bit. :-)

Dockerfile
Artanicus: This was imported as-is from ajacobson/chiadog-docker so @ajacobson might have more insight if there's some magic to it, but from what I can see it should not be required assuming we can install the binary wheel. Since the base is just `ubuntu:latest` I don't see why that wouldn't work. Will simplify & test.

Dockerfile
Artanicus: This was imported as-is, I'd assume the intent was to have a consistent UTC TZ when none is provided. But, since UTC is already the unconfigured default this shouldn't be necessary. Will simplify & test.

Dockerfile
Artanicus: Actually we don't need to install pynacl separately at all since the requirements.txt will pull it in straight to the venv. So this gets simplified quite a bit.

Dockerfile
Artanicus: Tests show we don't really need to do any of this mucking, just having the right TZ in the env is enough to get chiadog on the right timezone. So I removed all of it.

Dockerfile
Artanicus: Fix'd. This is a potentially breaking incompatibility to anyone running the original image, but since it's a manual action to switch to the official one, not a big deal.

Artanicus:
@martomi PTAL, all feedback incorporated and everything simplified quite a bit.

martomi (approved):
Looks great now, thanks!

PR #331, opened 2022-03-14
Add a transaction amount filter

egeedot:
Adds a transaction filter which helps deal with recent frequent Chia dust storms. Particularly important when using integrations such as Pushover which have a monthly cap (which I have already hit due to the excessive amount of messages).

A test has also been added for the new functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ helps with automated monitoring and sends you a mobile notification in case some
 | ------------- | ------------- | ------|
 | Harvester | Your harvester appears to be offline! No events for the past 400 seconds. | HIGH |
 | Harvester | Disconnected HDD? The total plot count decreased from 100 to 40. | HIGH |
+| Harvester | Connected HDD? The total plot count increased from 0 to 42. | LOW |
 | Harvester | Experiencing networking issues? Harvester did not participate in any challenge for 120 seconds. It's now working again. | NORMAL |
 | Harvester | Seeking plots took too long: 21.42 seconds! | NORMAL |
 | Full Node | Experiencing networking issues? Skipped 42 signage points! | NORMAL |
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ You may use one (or more) of the following integrations to receive notifications
 | Integration | Advantages | Cost |
 | ------------- | ------------- | ------|
 | [Pushover](https://pushover.net/) | High priority notifications that can override your phone's silent mode. | $5 one time purchase after 30 day trial. |
+| [Pushcut](https://pushcut.io/) | Alternative to Pushover |
 | E-mail | You probably already have an email. No additional apps. | Free |
 | Slack | Quick & easy setup. | Free |
 | Discord | Quick & easy setup. | Free |
--- a/README.md
+++ b/README.md
@@ -196,16 +196,19 @@ nohup python3 -u main.py --config config.yaml > output.log 2>&1 &
 To stop chiadog, you can find the Process ID (PID) via `ps aux | grep main.py` and then softly interrupt the process
 with `kill -SIGINT <pid_here>`.
 
+You can also run chiadog as a sandboxed systemd service.
+
 ## Running `chiadog` in sandboxed environment
 
-We're in the early stages of exploring the best way to provide easy to setup sandboxed environment where the `chiadog`
-process is completely isolated from potentially accessing your private keys. Contributions in that direction are very
-welcome. Meanwhile you can check out @ajacobson repository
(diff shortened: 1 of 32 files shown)


tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -34,6 +34,15 @@ def testFloatPrecision(self):
...
         self.assertEqual(events[0].message, "Cha-ching! Just received 0.000000000001 XCH ☘️")
 
+    def testTransactionAmountFilter(self):
+        filter_handler = WalletAddedCoinHandler(config={
+            'transaction_amount': "0.000000000001"
+        })
+        with open(self.example_logs_path / "small_values.txt", encoding="UTF-8") as f:
+            logs = f.readlines()
+        events = filter_handler.handle("".join(logs))
+        self.assertEqual(0, len(events))
```
martomi, line 44: suggestion: let's add at least 1 value above the threshold in the example file, so we can hit both below/above threshold in the same test?

README.md
```diff
@@ -209,6 +209,21 @@ user and blocks access to key chia locations.
...
 [chiadog-docker](https://github.com/ajacobson/chiadog-docker).
+
+## Filters
+
+Filters are also supported by adding them under the `filter` section, i.e.
+
+```yaml
+filter:
+  transaction_amount: "0.000000000002"
+```
```
martomi, line 220: nice-to-have: Thinking about future extensibility of this, I think better structure would be:

```
handlers:
    wallet_added_coin_handler:
        filters:
            transaction_amount:  "0.000000000002"
```

then we can also use the same structure in future to allow selective enable / disable of filters like:

```
handlers:
    wallet_added_coin_handler:
        enable: False
        filters:
            transaction_amount:  "0.000000000002"
```

Important to get it as general as possible from the start, because adding fields later is much easier than migrations and backward compatibility for existing users. What do you think?

README.md
```diff
@@ -209,6 +209,21 @@ user and blocks access to key chia locations.
...
+
+The above example will filter out all chia messages for transactions that are not bigger than `0.000000000002`. This is
+useful in reducing spam that is often a result of chia
```
martomi, line 223: Thanks for documenting this. We typically document anything config related directly inside the [config-example.yaml](https://github.com/martomi/chiadog/blob/main/config-example.yaml). Please add the new configs there and feel free to create a multiline comment describing the purpose and function of the setting.

The benefit is that users can ideally understand all possible configs by only looking at the example config and not cross-referencing other parts of the docs.

martomi (changes requested):
Great work, thanks for contributing! 👍 Overall looks good but I had few ideas how to make it even better, please have a look :-) Most importantly, one test file seems to be missing.

Off-topic: Keep in mind it's also possible to disable `wallet_events` for `pushover` in the config. (useful to know before this gets released)

egeedot:
@martomi I will do your changes this week.

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -34,6 +34,15 @@ def testFloatPrecision(self):
         self.assertEqual(events[0].service, EventService.WALLET, "Unexpected service")
         self.assertEqual(events[0].message, "Cha-ching! Just received 0.000000000001 XCH ☘️")
 
+    def testTransactionAmountFilter(self):
+        filter_handler = WalletAddedCoinHandler(config={
+            'transaction_amount': "0.000000000001"
+        })
+        with open(self.example_logs_path / "small_values.txt", encoding="UTF-8") as f:
```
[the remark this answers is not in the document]
egeedot: I reused an already existing `small_values.txt` file which what used in another test.

src/chia_log/handlers/wallet_added_coin_handler.py
```diff
@@ -14,8 +14,19 @@ class WalletAddedCoinHandler(LogHandler):
     receiving XCH and creates user notifications.
     """
 
-    def __init__(self):
+    def __init__(self, config: dict):
```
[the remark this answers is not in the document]
egeedot: I didn't end up doing this because I restructured the config as requested.

egeedot:
@martomi Completed everything as requested. I can also rebase the pull request if you want a single commit

martomi (changes requested):
Thanks for taking the time to implement the feedback. 🙌 

I spotted just one more semantic issue in the config structure that needs to be addressed. The other suggestion is optional.

martomi:
I was waiting on the last piece of feedback to be addressed. But considering a decent amount of time has passed already, let me address that myself so we can merge it.

Issue #340, opened 2022-07-02
Since Chia 1.4.0 received is always 0 XCH


tippiecat:
Since Chia 1.4.0 received is always 0 XCH. I upgraded on 6/30 and although I'm getting pooling rewards, ChiaDog is not reporting these anymore. This includes daily report and wallet receipt.

martomi:
Should be fixed now, thanks to @djerfy !

PR #341, opened 2022-07-03
Improve wallet parser for chia v1.4.0+

djerfy:
Improve wallet parser for Chia 1.4.0+ (tested). Related with issue #340.


martomi:
Thanks for the fix 🙌 

Please add some examples for the new log format in [this folder](https://github.com/martomi/chiadog/tree/main/tests/chia_log/logs/wallet_added_coin). I suggest to name the files `nominal-before-1.4.0.txt` and `nominal-after-1.4.0.txt` for example.

And a test case reading from the new log format [here](https://github.com/martomi/chiadog/blob/main/tests/chia_log/handlers/test_wallet_added_coin_handler.py).

martomi (changes requested):
Requesting changes as per comment above.

djerfy:
Thanks for your comment. Tests have been added and tested for me (with success and error) 👍

martomi:
Love it, thanks for the quick iteration!

martomi (approved):

PR #348, opened 2022-07-26
Specify image arch's to build

lryanuk:
Specify image arch's to include support for linux/arm64

Added using example documented here: https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -16,6 +16,12 @@ jobs:
       packages: write
 
     steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Checkout repository
         uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-image.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
           context: .
+          platforms: linux/amd64,linux/arm64
```
martomi, line 39: I'm pretty certain this will also require

```
 -
        name: Set up QEMU
        uses: docker/setup-qemu-action@v2
-
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v2
```
Can you please also do a test run of the pipeline in your repository and link to the successful result?

martomi (changes requested):

.github/workflows/publish-image.yml
lryanuk: Added the parts I missed, here's the successful build: https://github.com/lryanuk/chiadog/actions/runs/2883959326

Ran the container on a Raspberry Pi4, seems to be working as expected

martomi: That's great, thanks! 🙌

martomi (approved):

PR #362, opened 2023-01-17
Omit debug logging of zero mojo transactions.

Artanicus:
Since the flow of transactions is constant, logging zero transactions makes debug logs very hard to read. Non-zero mojo transactions under the set limit are still debug logged.

### DIFF
--- a/src/chia_log/handlers/wallet_added_coin_handler.py
+++ b/src/chia_log/handlers/wallet_added_coin_handler.py
@@ -47,7 +47,7 @@ class WalletAddedCoinHandler(LogHandlerInterface):
                     message=f"Cha-ching! Just received {xch_string} XCH ☘️",
                 )
             )
-        else:
+        elif total_mojos != 0:
             logging.debug(
                 f"Filtering out chia coin message since the amount ${total_mojos} received is less than"
                 f"the configured transaction_amount: ${self.min_mojos_amount}"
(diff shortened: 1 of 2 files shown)


martomi:
👋  Thanks for submitting this!

My understanding is this only logs transactions to your personal wallet. I don’t have a wallet running at the moment to check but is it really typical for wallets to receive a constant stream of 0 transactions or is this some targeted attack on specific wallets that started more recently? Could you give an estimate for the volume of 0 transactions you’re seeing?

Before we can merge, we also need to fix the failing CI check. It is failing on the `mypy` validation. You can reproduce this locally by running [these commands](https://github.com/martomi/chiadog/blob/main/CONTRIBUTING.md#formatting-and-linting). 

The error is not caused by this PR, I’m not sure how it passed before, but we’ll need to fix it to unblock merging. Basically you just need to replace line 27 in the `keep_alive_monitor.py` with:

```
def __init__(self, config: Optional[dict] = None, thresholds: Optional[dict] = None):
```

and add the import `Optional` from `typing` at the top.

Artanicus:
Looking at the logic closer, I believe the root cause is actually a logic error. `total_mojos = 0` initializes the value as zero, and then subsequent coin messages increment that value if they were received in the update. Ergo, if no coin messages were received at all, the value is still zero and the log message is logged.

Ergo, as long as `WalletAddedCoinHandler.handle()` gets triggered, a debug log will trigger for the default zero coin value unless they are filtered out silently.

martomi:
Ah, I see - yes, thats was a logical error indeed! 

Thanks for fixing the types.

martomi (approved):

Issue #361, opened 2023-01-16
Disabling harvester_activity_handler does not disable harvester checks


Artanicus:
Even if the handler `harvester_activity_handler` is set to `enabled: false`, a harvester is still expected: `Your harvester appears to be offline! No events for the past 600 seconds.`
- For context, this instance of chiadog is monitoring a full_node that has been started with `services="node farmer-only wallet"`, i.e. it's not running a harvester on purpose since there are no local plots.
- Issue repeats with Chia log level set to `DEBUG` or `INFO`, have not tested other levels.

```
[2023-01-16 15:29:28] [    INFO] --- Starting Chiadog (unknown) (main.py:54)
[2023-01-16 15:29:28] [    INFO] --- Enabled local file log consumer. (log_consumer.py:60)
[2023-01-16 15:29:28] [    INFO] --- Using temporary directory /tmp/tmpwt8e5db5/debug.log.offset (log_consumer.py:63)
[2023-01-16 15:29:28] [    INFO] --- Enabled remote pinging to https://hc-ping.com/xxxxx (keep_alive_monitor.py:36)
[2023-01-16 15:29:28] [    INFO] --- Keep-alive check period: 300 seconds (keep_alive_monitor.py:45)
[2023-01-16 15:29:28] [    INFO] --- Initializing Pushover notifier. (pushover_notifier.py:13)
[2023-01-16 15:29:28] [    INFO] --- Enabled stats for daily notifications (stats_manager.py:48)
[2023-01-16 15:29:28] [    INFO] --- Summary notifications will be sent out every 24 hours starting from 12:00 (stats_manager.py:61)
[2023-01-16 15:29:28] [    INFO] --- Disabled handler: harvester_activity_handler (log_handler.py:63)
[2023-01-16 15:29:28] [    INFO] --- Initializing handler: partial_handler (__init__.py:28)
[2023-01-16 15:29:28] [    INFO] --- Enabled parser for partial submitting stats. (partial_parser.py:28)
[2023-01-16 15:29:28] [    INFO] --- Initializing handler: block_handler (__init__.py:28)
[2023-01-16 15:29:28] [    INFO] --- Enabled parser for block found stats. (block_parser.py:28)
[2023-01-16 15:29:28] [    INFO] --- Enabled check for found blocks. (found_blocks.py:15)
[2023-01-16 15:29:28] [    INFO] --- Initializing handler: finished_signage_point_handler (__init__.py:28)
[...]

PR #363, opened 2023-02-03
Migrate configuration handling to the confuse lib

Artanicus:
This is very much a work in progress, just wanted to be transparent and make the current status visible. The current state is able to run the example config and known minimal configs but needs more testing.

This change touches all integrations and has minor potential for subtle breakage. The following have been thoroughly tested:
- Telegram
- Pushover

Groundwork for fixing #361


src/chia_log/handlers/daily_stats/stats_manager.py
```diff
@@ -36,10 +39,10 @@ class StatsManager:
     with a summary from all stats that have been collected for the past 24 hours.
     """
 
-    def __init__(self, config: dict, notify_manager: NotifyManager):
-        self._enable = config.get("enable", False)
-        self._notify_time = self._parse_notify_time(config.get("time_of_day", "21:00"))
-        self._frequency_hours = config.get("frequency_hours", 24)
+    def __init__(self, config: confuse.core.Configuration, notify_manager: NotifyManager):
+        self._enable = config["enable"].get()
```
martomi, line 43: Would be cool to make use of the [validation](https://confuse.readthedocs.io/en/latest/usage.html#validation) that `confuse` provides. Here maybe `get(bool)`.

src/chia_log/handlers/wallet_added_coin_handler.py
```diff
@@ -18,11 +21,10 @@ class WalletAddedCoinHandler(LogHandlerInterface):
     def config_name() -> str:
         return "wallet_added_coin_handler"
 
-    def __init__(self, config: Optional[dict] = None):
+    def __init__(self, config: confuse.core.Configuration):
```
martomi, line 24: This is rather a `confuse.core.ConfigView` right? I wonder how we can make it more readable without having to parse all the call-sites to determine the full nesting.

src/default_config.yaml
```diff
@@ -0,0 +1,138 @@
+# This file defines the default values used if not overriden in config.yaml
+
```
martomi, line 2: Let’s add some warning comment here to make sure users don’t modify this file, rather point them to the correct place for their custom config.yaml.

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -3,24 +3,28 @@
...
 
 
 class TestWalledAddedCoinHandler(unittest.TestCase):
     def setUp(self) -> None:
         config_dir = Path(__file__).resolve().parents[3]
-        config = Config(config_dir / "config-example.yaml")
-        self.handler_config = config.get_handlers_config()[WalletAddedCoinHandler.config_name()]
+        config = confuse.Configuration("chiadog", __name__)
+        config.set_file(config_dir / "src/default_config.yaml")
+        config.set_file(config_dir / "config-example.yaml")
```
martomi, line 19: I think we don’t need to second override for tests.

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -3,24 +3,28 @@
...
+        self.handler_config = config["handlers"][WalletAddedCoinHandler.config_name()]
```
martomi, line 20: Tied to another comment, maybe the handler class can have a method that returns the fully materialized view from the rootview. Then the code here would be slightly cleaner:
```
self.config_view_handler = WalletAddedCoinHandler.config_view(root_view)
```

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -40,7 +44,12 @@ def testFloatPrecision(self):
         with open(self.example_logs_path / "small_values.txt", encoding="UTF-8") as f:
             logs = f.readlines()
 
-        events = self.handler.handle("".join(logs))
+        default_config = self.handler_config
+        no_filter_config = copy.deepcopy(default_config)
+        no_filter_config["min_mojos_amount"] = 0
+        no_filter_handler = WalletAddedCoinHandler(no_filter_config)
```
martomi, line 50: I think you can do `self.handler_config[“min_mojos_amount”] = 0` to directly override without copies. And then add a `self.handler_config.clear()` cleanup hook before or after each test.

Artanicus:
That all makes sense, thanks for the thorough review!

Artanicus:
I think this might now be in a sane state, but was only able to test a couple of the notifiers myself. @martomi if you have some test farm for all the notifiers, could you give it a whirl and make sure I didn't miss anything?

src/notifier/__init__.py
```diff
@@ -60,17 +63,17 @@ class Notifier(ABC):
...
-    def __init__(self, title_prefix: str, config: dict):
+    def __init__(self, title_prefix: str, config: ConfigView):
         self._title_prefix = title_prefix
         self._config = config
         self._conn_timeout_seconds = 10
         self._notification_types = [EventType.USER]
         self._notification_services = [EventService.HARVESTER, EventService.FARMER, EventService.FULL_NODE]
 
-        daily_stats = config.get("daily_stats", False)
-        wallet_events = config.get("wallet_events", False)
```
martomi, line 71: Can we once again go through the defaults for all values and make sure we’re keeping them the same please?

Here `daily_stats` used to have been disabled by default. But the new `default_config.yaml` has it enabled. Same thing applies for `wallet_events`.

I’d like to keep the migration to `confuse` separate of any decisions on changing the existing defaults as this will require more nuanced discussions on backward compatability and will delay this PR. Concrete example: there are users with multiple notifiers with legacy configs that don’t override the `wallet_event` values. They might start getting overwhelming number of notifications from pool rewards if this gets merged.

src/notifier/discord_notifier.py
```diff
@@ -4,16 +4,19 @@
...
 
 
 class DiscordNotifier(Notifier):
-    def __init__(self, title_prefix: str, config: dict):
+    def __init__(self, title_prefix: str, config: ConfigView):
         logging.info("Initializing Discord notifier.")
         super().__init__(title_prefix, config)
         try:
-            credentials = config["credentials"]
+            credentials = config["credentials"].get(dict)
```
martomi, line 19: Unfortunately, I don’t have a setup to run more comprehensive tests on the notifiers. Were you able to successfully run one of them which included this `get(dict)` parsing to make sure credentials are generally getting parsed in backward compatible way with the new library?

src/notifier/mqtt_notifier.py
```diff
@@ -30,9 +33,9 @@ def _set_config(self, config: dict):
         :param config: The YAML config for this notifier
         :returns: None
         """
-        self._topic = config["topic"]
-        self._qos: int = config.get("qos", 0)
-        self._retain: bool = config.get("retain", False)
```
martomi, line 35: I’m not making an exhaustive list of mismatches but here I also noticed `qos` used to be 0, and it’s 1 now.

src/notifier/slack_notifier.py
```diff
@@ -5,16 +5,19 @@
...
 
 
 class SlackNotifier(Notifier):
-    def __init__(self, title_prefix: str, config: dict):
+    def __init__(self, title_prefix: str, config: ConfigView):
         logging.info("Initializing Slack notifier.")
         super().__init__(title_prefix, config)
         try:
-            credentials = config["credentials"]
+            credentials = config["credentials"].get()
```
martomi, line 20: `get(dict)` to be consistent with others

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -3,24 +3,31 @@
...
+        self.config = confuse.Configuration("chiadog", __name__)
+        self.config.set_file(config_dir / "src/default_config.yaml")
+        self.handler_config = self.config["handlers"][WalletAddedCoinHandler.config_name()]
 
-        self.handler = WalletAddedCoinHandler(config=None)
+        self.handler = WalletAddedCoinHandler(config=self.handler_config)
         self.example_logs_path = Path(__file__).resolve().parents[1] / "logs/wallet_added_coin"
 
+    def tearDown(self) -> None:
+        # self.handler_config.clear()
```
martomi, line 25: Maybe you wanted to either delete or uncomment?

tests/chia_log/handlers/test_wallet_added_coin_handler.py
```diff
@@ -40,7 +47,11 @@ def testFloatPrecision(self):
         with open(self.example_logs_path / "small_values.txt", encoding="UTF-8") as f:
             logs = f.readlines()
 
-        events = self.handler.handle("".join(logs))
+        # Default mojo filter excludes the event we will expect
+        self.handler_config["min_mojos_amount"].set(0)
+        handler = WalletAddedCoinHandler(self.handler_config)
```
martomi, line 52: Since the default used to be 0, maybe you want to do it the other way around? Keep it 0 and set it to 5 only for the test that requires to test that filter’s functionality?

tests/notifier/test_ifttt_notifier.py
```diff
@@ -46,15 +53,15 @@ def testShowcaseGoodNotifications(self):
         notifiers = [
             IftttNotifier(
                 title_prefix="Harvester 1",
-                config={"enable": True, "api_token": self.api_token, "webhook_name": self.webhook_name},
+                config=self.config,
```
martomi, line 56: This config used to be the wrong format before, right?

martomi (changes requested):
Looks good on a high-level. Left a bunch of small comments.

Agree with some of the default value changes but would like to keep that separated for another potential PR in the future.

Can you please also document in the PR description which integrations you managed to test? Since this change is touching all integrations, I’d like us to test at least 3 of them (feel free to pick which ones). The guide for testing is [here](https://github.com/martomi/chiadog/blob/main/INTEGRATIONS.md). Thanks!

src/chia_log/handlers/daily_stats/stat_accumulators/search_time_stats.py
```diff
@@ -22,14 +22,13 @@ def reset(self):
 
     def consume(self, obj: HarvesterActivityMessage):
         self._num_measurements += 1
-        self._avg_time_seconds += (obj.search_time_seconds - self._avg_time_seconds) / self._num_measurements
+        self._avg_time_seconds += int((obj.search_time_seconds - self._avg_time_seconds) / self._num_measurements)
```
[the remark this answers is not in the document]
Artanicus: This was causing typing warnings and I think it was implicitly getting dropped to int anyway. Let me double check if we can do it as a float instead.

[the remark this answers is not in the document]
Artanicus: Yeah, I just solved the typing silly. Fixed the init as an int instead and now it's happy.

src/chia_log/handlers/condition_checkers/non_skipped_signage_points.py
```diff
@@ -17,7 +18,7 @@ class NonSkippedSignagePoints(FinishedSignageConditionChecker):
 
     def __init__(self):
         logging.info("Enabled check for finished signage points.")
-        self._last_signage_point_timestamp = None
+        self._last_signage_point_timestamp = datetime.fromtimestamp(0)
```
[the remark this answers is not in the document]
Artanicus: The `check()` actually relies on the signage_point, not it's timestamp. But, I gave the signage point typing and changed the check value to `0` to make it clearer.

src/chia_log/handlers/daily_stats/stat_accumulators/signage_point_stats.py
```diff
@@ -9,7 +9,7 @@
 class SignagePointStats(FinishedSignageConsumer, StatAccumulator):
     def __init__(self):
         self._last_reset_time = datetime.now()
-        self._last_signage_point_timestamp = None
+        self._last_signage_point_timestamp = datetime.fromtimestamp(0)
```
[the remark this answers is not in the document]
Artanicus: Fixed it the same way.

src/chia_log/log_consumer.py
```diff
@@ -16,12 +16,12 @@
...
 
 # project
-from src.config import Config
-from src.config import check_keys
 from src.util import OS
 
 # lib
 import paramiko
+from confuse import ConfigView
+from confuse import Path as confuse_Path
```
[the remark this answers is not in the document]
Artanicus: Doh, this was a total brainfart. Good catch.

src/notifier/mqtt_notifier.py
Artanicus: Yeah, so what we have here is that I took the default values from config-example.yaml instead of from the code, with the assumption that anyone actually using it was likely always defaulting on the example instead of the default in the code.

So I think what we have here is a problem of having had two defaults and we need to collapse on only one. You wana do that one by one or should I just flip over from example config values to code defaults?

src/notifier/slack_notifier.py
Artanicus: Yup, good catch.

tests/chia_log/handlers/test_wallet_added_coin_handler.py
Artanicus: Yup, good catch. SubViews can't be cleared so it was useless.

tests/chia_log/handlers/test_wallet_added_coin_handler.py
Artanicus: That makes sense, but I guess we need to first agree if the default should be 5 or 0 since it's a conflict between the code default and the example config.

tests/notifier/test_ifttt_notifier.py
Artanicus: I think so, or perhaps not. The config structure on this one is really weird with the webhook_name under credentials. I'll go whip up an IFTTT account and test this one for real :D

Artanicus:
I think I've fixed the ones where there was a clear fix. Happy to tidy up the default values if we have a consensus on which source of truth should be the new defaults. My feeling is that it's unlikely many folks have relied on the code defaults and instead have used the example values instead of explicitly deleting those fields from their config. But, alas I have no way of proving that feeling!

I've got the keepalive structural changes and a keepalive emitter for the wallet ready for review once we nail down this bit, just want to rebase them first on top of this one so you get a clean review slate.

martomi:
Cool - takes for the quick turnaround!

Note that there are a few more inline comments I left in the code above, but they’re collapsed in the middle of the thread by the github UI. Just wanna make sure you saw? In one of them I had provided more context and example for default values.

My take is that we should use the code defaults for the following reasons:
- The example config is still going to be taken by new users, so the default YAML has little effect beside keeping the code functionality being identical as how it used to be before this change
- There were a substantial number of users that installed `chiadog` before we introduced most of the configs in the example config. In the process of installing it, they would have created a copy of the very old config example. We never explicitly required them to change their original configs with the new values from the extended example config. Thus they’ll be having a config that is substantially more minimal than the current example. By keeping the defaults same as in the code, we’ll not introduce any unexpected changes for these users.

src/chia_log/log_consumer.py
```diff
@@ -217,40 +216,38 @@ def create_log_consumer_from_config(config: dict) -> Optional[LogConsumer]:
     enabled_consumer_config = config[enabled_consumer]
 
     if enabled_consumer == "file_log_consumer":
-        if not check_keys(required_keys=["file_path"], config=enabled_consumer_config):
+        if "file_path" not in enabled_consumer_config or not enabled_consumer_config["file_path"]:
```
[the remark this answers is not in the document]
Artanicus: I believe it's actually impossible to fail in that way anymore, since not defining file_path will fall back to the default, i.e. `~/.chia/mainnet/log/debug.log`. The only way I found is if you provide an empty `chia_logs` block, which crashes and burns.

tests/notifier/test_ifttt_notifier.py
Artanicus: Can confirm that on branch main, if `SHOWCASE_NOTIFICATIONS=true` the test fails with a bunch of:
> ERROR:root:Invalid config.yaml. Missing key: 'credentials'

And eventually

> Traceback (most recent call last):
  File "/home/artanicus/src/3rd-party-forks/chiadog.git/tests/notifier/test_ifttt_notifier.py", line 67, in testShowcaseGoodNotificat
ions
    success = notifier.send_events_to_user(events=[found_proof_event])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/artanicus/src/3rd-party-forks/chiadog.git/src/notifier/ifttt_notifier.py", line 30, in send_events_to_user
    f"/trigger/{self.webhook_name}/json/with/key/{self.token}",
                ^^^^^^^^^^^^^^^^^ 
AttributeError: 'IftttNotifier' object has no attribute 'webhook_name'

Ergo, the config indeed has not been valid. It was probably never caught since these lines are not triggered unless the showcase is enabled. With the new code the showcases work as expected.

Artanicus:
Oops, not sure why "close" is on by default in review mode. But did not intend to close and now can't reopen :D

martomi:
Hmm not sure, the reopen button is also grayed out for me. But more importantly, the code-diff is missing, so maybe that’s the main reason? Maybe you rebased on the wrong base?

Artanicus:
Yeah, looks like there's no sane way to re-open after a rebase & close. I'll open a new PR.

PR #364, opened 2023-02-10
Migrate configuration handling to the confuse lib, take 2

Artanicus:
This enables:
- Conservative config defaults in `src/default_config.yaml`, which leads to smaller and leaner custom configs.
- Config validation with confuse templates or at-use get methods with type checking.
- Easier future mutation of config format as defaults can carry over sane defaults.

This change touches all integrations and has minor potential for subtle breakage. The following have been thoroughly tested:

- Telegram
- Pushover
- IFTTT
- MQTT
- SMTP

Notifiers not yet tested:
- Discord
- Slack
- Script

Notifiers with known issues:
- Grafana, WIP

Notifiers I cannot test myself:
- Pushcut

Groundwork for fixing #361

PR #365, opened 2023-02-11
Fall back to SMTP without auth if server doesn't support it

Artanicus:
As the default is to enable auth, a server without support produces a very long stracktrace. With this change instead a clear error log is emitted and the connection is still attempted without auth.

### DIFF
--- a/src/notifier/smtp_notifier.py
+++ b/src/notifier/smtp_notifier.py
@@ -70,7 +70,13 @@ class SMTPNotifier(Notifier):
                     # stmplib docs recommend calling ehlo() before & after starttls()
                     server.ehlo()
                     if self.enable_smtp_auth:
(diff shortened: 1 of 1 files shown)


Artanicus:
This PR includes a bunch of other minor changes that overlap with the config-revamp PR #364. This is due to the typing issues already being present in the current HEAD, so likely issues found by newer versions of mypy.

Artanicus:
Actually, the only thing I truly *had* to update was the `black` run & `flake8` warning. The typing errors I was seeing were due to including the `--check-untyped-defs` mypy parameter. So if desired, I can roll this back to just b417874
& 4b3af01. But those typing fixes are in the other PR as well.

martomi (approved):
Looks good, thanks for making it fail more gracefully.

PR #366, opened 2023-02-12
Fix up the Grafana notifier so it's more standard and functions post config-revamp

Artanicus:
While not all options actually make sense for it, it's easier to provide them than special case just because no one would actually ever enable them.

NOTE: This does not mean the Grafana notifier actually supports any more events than it does right now. It probably should, but that's out of scope.

@pieterhelsen As the original author of the integration. I'm not actually managing to push the annotations to Grafana no matter how much I try, so could you check that I'm not accidentally changing anything critical here? For example, code comments lead me to believe null would default to global annotations, but `-1` is the example value given, so not sure what the truth there is. Also the test doesn't yet support defining those values.

### DIFF
--- a/src/default_config.yaml
+++ b/src/default_config.yaml
@@ -100,12 +100,12 @@ notifier:
       username: null
       password: null
   grafana:
-    enable: false
+    <<: *notifier_defaults
     credentials:
       base_url: null
       api_token: null
-      dashboard_id: -1
-      panel_id: -1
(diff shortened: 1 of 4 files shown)


martomi (approved):
I’ll go ahead and merge this optimistically and hope it achieves the fix.

@pieterhelsen if you manage to test it please let us know :)

PR #367, opened 2023-02-12
Enable CI checks also for main.py

Artanicus:
And fix a few things brought up by black and mypy.

### DIFF
--- a/main.py
+++ b/main.py
@@ -73,9 +73,7 @@ def init(config: confuse.core.Configuration):
 
     # Link stuff up in the log handler
     # Pipeline: Consume -> Handle -> Notify
-    log_handler = LogHandler(
-        config=config, log_consumer=log_consumer, notify_manager=notify_manager, stats_manager=stats_manager
-    )
+    LogHandler(config=config, log_consumer=log_consumer, notify_manager=notify_manager, stats_manager=stats_manager)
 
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -57,25 +56,26 @@ def init(config: confuse.core.Configuration):
     logging.info(f"Starting Chiadog ({version()})")
 
     # Create log consumer based on provided configuration
-    chia_logs_config = config['chia_logs']
-    log_consumer = create_log_consumer_from_config(chia_logs_config)
+    chia_logs_config = config["chia_logs"]
+    log_consumer: LogConsumer = create_log_consumer_from_config(chia_logs_config)
     if log_consumer is None:
         exit(0)
```
martomi, line 62: nit: this check isn’t needed anymore.

Artanicus: Fixed.

Artanicus:
Fixed the nit and threw in flake8 testing for main.py as well.

main.py
martomi: ah, here I actually meant that we don’t need the if-statement since the function doesn’t anymore return `None`

martomi: But its a nit, happy to merge as-is

martomi (approved):

PR #368, opened 2023-02-12
Groundwork for keepalives for all services

Artanicus:
This PR deprecates enabling or disabling individual handlers in favor of service level enablement, which in turn enables all handlers defined for that service.

More ground work for fixing #361. At this PR it's already possible to disable harvester monitoring but there are no other keepalives available yet. The next PR in the chain adds a new keepalive for the wallet as an example.

I'm torn on which services should be monitored by default. The conservative default would be no services, but that clashes with the previous default of running all handlers and the harvester keepalive. So I've opted to default to monitoring all services and thus all handlers, which also enables the harvester keepalive. Where it gets a bit wonky then is when we bring in a wallet keepalive. There's been no harm in having the incoming transaction handler enabled by default, but by comparison the peak handler expects to find a working wallet. If one isn't found, it will start triggering keepalive alerts of no wallet activity.

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -27,6 +27,15 @@ chia_logs:
 keep_alive_monitor:
   enable_remote_ping: false
   ping_url: ''
+  # These thresholds determine how long a service can be unhealthy,
+  # before we trigger a high priority alert.
+  # The lowest value here determines how often the keep-alive checks run,
+  # and thus how often a notification for an unhealthy service will trigger!
+  notify_threshold_seconds:
+    FULL_NODE: 300
+    HARVESTER: 300
+    FARMER: 300
+    WALLET: 300
 
 # Enable this and you'll receive a daily summary notification
 # on your farm performance at the specified time of the day.
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -35,26 +35,20 @@ daily_stats:
   time_of_day: "21:00"
   frequency_hours: 24
 
-# In this section you can enable and configure the log handlers
+# Remove any service your node isn't running. All are enabled by default.
+# Services listed here are checked for health and will start alerting if missing.
+monitored_services:
+  - FULL_NODE
+  - HARVESTER
+  - FARMER
+  - WALLET
+
+# In this section you can configure log handler details
 handlers:
-  # Checks for new coins received (wallet)
   wallet_added_coin_handler:
-    enable: true
     # Transactions with lower amount mojos will be ignored
     # Use this to avoid notification spam during dust storms
     min_mojos_amount:  5 # default: 0
-  # Checks for skipped signage points (full node)
-  finished_signage_point_handler:
-    enable: true
-  # Checks for new blocks found (full node)
-  block_handler:
-    enable: true
-  # Checks for found partials (farmer)
-  partial_handler:
-    enable: true
-  # Checks harvester activity & health (harvester)
-  harvester_activity_handler:
-    enable: true
 
 
 # We support a lot of notifiers, please check the README for more
--- a/src/chia_log/log_handler.py
+++ b/src/chia_log/log_handler.py
@@ -51,10 +51,11 @@ class LogHandler(LogConsumerSubscriber):
         self._active_handlers = []
         for service, service_handlers in self.services.items():
             if service.name in config["monitored_services"].get(list):
+                logging.info(f"Enabled service monitoring: {service.name}")
                 for handler in service_handlers:
                     self._active_handlers.append(handler(config["handlers"][handler.config_name()]))
             else:
-                logging.info(f"Disabled service monitoring: {service.name}")
+                logging.debug(f"Disabled service monitoring: {service.name}")
         log_consumer.subscribe(self)
 
     def consume_logs(self, logs: str):
--- a/src/chia_log/log_handler.py
+++ b/src/chia_log/log_handler.py
@@ -49,16 +49,12 @@ class LogHandler(LogConsumerSubscriber):
         self._stats_manager = stats_manager
 
         self._active_handlers = []
-        self._active_services = []
         for service, service_handlers in self.services.items():
             if service.name in config["monitored_services"].get(list):
-                self._active_services.append(service)
                 for handler in service_handlers:
                     self._active_handlers.append(handler(config["handlers"][handler.config_name()]))
             else:
                 logging.info(f"Disabled service monitoring: {service.name}")
-        # Init the keepalive monitor with enabled services
-        notify_manager._keep_alive_monitor.set_services(self._active_services)
         log_consumer.subscribe(self)
 
     def consume_logs(self, logs: str):
--- a/src/chia_log/log_handler.py
+++ b/src/chia_log/log_handler.py
@@ -1,10 +1,9 @@
 # std
-from typing import Optional, List, Type
+from typing import Optional, List, Type, Dict
 import logging
 
 # lib
 from confuse import ConfigView
-from confuse.exceptions import ConfigTypeError
 
 # project
 from src.chia_log.handlers import LogHandlerInterface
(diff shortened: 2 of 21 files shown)


Artanicus:
The sneak peek of the wallet peak handler and wallet keepalive:
https://github.com/Artanicus/chiadog/compare/keepalive...Artanicus:chiadog:keepalive_wallet?expand=1

src/chia_log/log_handler.py
```diff
@@ -48,26 +39,29 @@ def __init__(
...
+            if service.name in config["monitored_services"].get(list):
+                self._active_services.append(service)
+                for handler in service_handlers:
+                    self._active_handlers.append(handler(config["handlers"][handler.config_name()]))
             else:
-                logging.info(f"Disabled handler: {handler.config_name()}")
-
+                logging.info(f"Disabled service monitoring: {service.name}")
+        # Init the keepalive monitor with enabled services
+        notify_manager._keep_alive_monitor.set_services(self._active_services)
```
martomi, line 61: The keep-alive monitor already has access to the config, I don’t see a good reason for coupling the initialization of the monitor with the log handler. 🤔 

Can we fully initialize the KeepAliveMonitor from its own `__init__` method with 
```
self.set_services(config["monitored_services"].get(list))
```

src/notifier/keep_alive_monitor.py
```diff
@@ -24,21 +27,45 @@ class KeepAliveMonitor:
...
-    def __init__(self, config: Optional[dict] = None, thresholds: Optional[dict] = None):
+    def __init__(self, config: ConfigView):
         self._notify_manager = None
+        self.config = config
 
-        self._last_keep_alive = {EventService.HARVESTER: datetime.now()}
-        self._last_keep_alive_threshold_seconds = thresholds or {EventService.HARVESTER: 300}
+        # These will be set by set_services() once known
+        self._last_keep_alive: Dict[EventService, datetime] = {}
+        self._last_keep_alive_threshold_seconds: Dict[EventService, int] = {}
```
martomi, line 36: As discussed in previous comment, `set_services` would ideally be called directly below these lines so we know for sure they’re being initialized.

src/notifier/keep_alive_monitor.py
```diff
@@ -24,21 +27,45 @@ class KeepAliveMonitor:
...
+            self._ping_url = self.config["ping_url"].get()
             logging.info(f"Enabled remote pinging to {self._ping_url}")
 
-        # Infer check period from minimum threshold (arbitrary decision)
+        # Check period will be inferred from minimum threshold of all services.
         # Note that this period defines how often high priority notifications
         # will be re-triggered so < 5 min is not recommended
         self._check_period = float("inf")
+        # No point in starting anything yet since we have no known services.
+        self._is_running = False
```
martomi, line 48: Probably won’t need that once we initialized services from within this method.

src/notifier/keep_alive_monitor.py
```diff
@@ -74,18 +97,19 @@ def check_last_keep_alive(self):
             events = []
             for service in self._last_keep_alive.keys():
                 seconds_since_last = (datetime.now() - self._last_keep_alive[service]).seconds
-                logging.debug(f"Keep-alive check for {service.name}: Last activity {seconds_since_last} seconds ago.")
-                if seconds_since_last > self._last_keep_alive_threshold_seconds[service]:
+                threshold = self._last_keep_alive_threshold_seconds[service]
+                logging.debug(f"Keep-alive check for {service}: Last activity {seconds_since_last}s/{threshold}s")
```
martomi, line 101: nit: Not sure that

> Last activity 200s/300s 

will be clear enough. Shall we make it a bit more verbose, e.g.

> Last activity 200s (notify threshold 300s)

src/notifier/keep_alive_monitor.py
```diff
@@ -74,18 +97,19 @@ def check_last_keep_alive(self):
...
+                if seconds_since_last > threshold:
                     message = (
-                        f"Your harvester appears to be offline! "
+                        f"Your {service.name} appears to be offline! "
                         f"No events for the past {seconds_since_last} seconds."
```
martomi, line 105: Here we should probably add some helpful hints, e.g.

> No events for the past 320 seconds. You can adjust your config.yaml to either increase the notify threshold from {current threshold}, or disable the checks for this service by removing {service.name} from the list of `monitored_services`.

martomi: But maybe this verbosity should only be included in the `logging.warning` below and not the actual notification.

src/default_config.yaml
```diff
@@ -18,28 +18,34 @@ chia_logs:
     remote_user: "chia"
     remote_port: 22
 
+# All services and thus handlers are enabled by default
+monitored_services:
+  - FULL_NODE
+  - HARVESTER
+  - FARMER
+  - WALLET
```
martomi, line 26: I think this should go in the example config for users, or are you keeping it for a follow-up PR?

martomi (changes requested):
Agreed with the default approach here. Few other comments / thoughts - PTAL.

src/notifier/keep_alive_monitor.py
martomi: I think it doesn’t hurt including in both.

src/chia_log/log_handler.py
Artanicus: I'll have another go at it. This is what I initially tried but somehow it got tricky that way as well.

src/notifier/keep_alive_monitor.py
Artanicus: I've also been thinking that "offline" might not be the best term here. We could instead phrase it as "not healthy" since offline has a strong connotation that the service isn't running, when in reality in most cases it's is sort of online, just wedged somehow. :D 

If it's rephrased as healthiness, then the messaging could be something like `Your {service.name} is unhealthy and hasn't behaved as expected for the past {seconds_since_last} seconds. (This check can be adjusted.)` .. which hopefully would lead anyone that is annoyed by it to go read documentation on how to not monitor that service, or to adjust the thresholds.

src/chia_log/log_handler.py
Artanicus: Found an elegant-ish solution to the initial problem by expanding `EventServices` to have a string init. This has the added benefit of ignoring capitalization of the service names in the config.

src/default_config.yaml
Artanicus: If this would be omitted, no handlers would get init by default. That's certainly a stance we can take, just not sure we're on the same page about the effects?

Artanicus: Oh now I get what you meant. Added also to example with more guidance.

src/notifier/keep_alive_monitor.py
Artanicus: Now it will be:
> [2023-02-17 15:44:32] [   DEBUG] --- Keep-alive check for EventService.HARVESTER: Last activity 5s (notify threshold 60s) (keep_alive_monitor.py:75)

src/notifier/keep_alive_monitor.py
Artanicus: Updated wording, LMK if that looks good.

martomi (changes requested):
Looks good, just a couple of seemingly duplicate lines. Also I’d like us to increase test coverage of the keep alive monitor now that were doing more work on it. Can also be a followup PR.

config-example.yaml
```diff
@@ -27,6 +27,15 @@ chia_logs:
 keep_alive_monitor:
   enable_remote_ping: false
   ping_url: ''
+  # These thresholds determine how long a service can be unhealthy,
+  # before we trigger a high priority alert.
+  # The lowest value here determines how often the keep-alive checks run,
+  # and thus how often a notification for an unhealthy service will trigger!
```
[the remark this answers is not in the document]
Artanicus: Yeah, that makes sense. Have a default and a setting, not too hard to implement after the config revamp.

src/notifier/keep_alive_monitor.py
Artanicus: Yeah, just my brainfart after the refactor. Fixd.

Artanicus:
Current code coverage overall with most notifiers enabled (8 tests skipped and Grafana failing) is 78%. Not bad, not great. Keep-alive is at 87%. Biggest misses are:
- Remote keep-alive isn't tested. I've previously used `mbtest` for these kinds of things but it's a really heavy testing dependency so will have a go at using `VCR.py` since it's way more self contained and faster.
- We don't test what happens without a `NotifyManager`, probably should.
- A couple of error conditions are not triggered, probably better to just exclude those from coverage.

Then there's actually testing novel situations, will have a think of those wile getting the pure coverage up first.

martomi (approved):
🚀

PR #371, opened 2023-02-16
@mention for high priority events in discord

WayneWayner:


src/notifier/discord_notifier.py
```diff
@@ -25,6 +25,8 @@ def send_events_to_user(self, events: List[Event]) -> bool:
         errors = False
         for event in events:
             if event.type in self._notification_types and event.service in self._notification_services:
+                if event.priority == EventPriority.HIGH:
+                    event.message += "\n@here"
```
martomi, line 29: Without testing this, I’m pretty sure this will modify the event in the list such that any subsequent notifiers will receive the modified message as well. In general I’d highly discourage modifying values that are passed in as a parameter - we should consider these immutable.

Please add this within a new `content` variable that can be provided in the request below.

martomi (changes requested):
Thanks, this seems useful in general. Needs slight rework - then I can merge it.

martomi (approved):

PR #372, opened 2023-02-18
Test remote pinging of keep_alive_monitor

Artanicus:
This bumps up keep_alive_monitor coverage to 100%.

introduces a new testing dependency: `vcrpy`. For convenience, testing_requirements.txt now contains all testing tools.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -47,3 +47,6 @@ jobs:
       with:
           coverageFile: coverage.xml
           token: ${{ secrets.GITHUB_TOKEN }}
+          thresholdAll: 0.6
+          thresholdNew: 0.9
+          thresholdModified: 0.7
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -43,7 +43,7 @@ jobs:
         python -m coverage run -m unittest
         python -m coverage xml coverage.xml
     - name: Check Coverage
-            uses: orgoro/coverage@v3
-            with:
-                coverageFile: coverage.xml
-                token: ${{ secrets.GITHUB_TOKEN }}
+      uses: orgoro/coverage@v3
+      with:
+          coverageFile: coverage.xml
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -25,9 +25,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install black flake8 mypy
-        pip install types-PyYAML types-python-dateutil types-paramiko types-retry
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        if [ -f testing_requirements.txt ]; then pip install -r testing_requirements.txt; fi
     - name: Format check with Black
       run: black --check src tests *.py
     - name: Lint with flake8
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -40,4 +40,10 @@ jobs:
         mypy --install-types --non-interactive --check-untyped-defs src tests *.py
     - name: Unit Tests
       run: |
-        python -m unittest
+        python -m coverage run -m unittest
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -25,9 +25,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install black flake8 mypy
-        pip install types-PyYAML types-python-dateutil types-paramiko types-retry
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        if [ -f testing_requirements.txt ]; then pip install -r testing_requirements.txt; fi
```
martomi, line 29: 👍 Could you please also reflect this change in https://github.com/martomi/chiadog/blob/main/CONTRIBUTING.md

martomi: Also does it make sense to add some CI test that prevents significant regressions on coverage?

tests/cassette/keep_alive_monitor_remote_ping.yaml
```diff
@@ -0,0 +1,66 @@
+interactions:
```
martomi, line 1: Could you add to the PR description what was the process to generate this yaml file just for later reference?

tests/notifier/test_keep_alive_monitor.py
```diff
@@ -68,20 +69,18 @@ def callback(events: List[Event]):
...
+                self.keep_alive_monitor.process_events(self.keep_alive_events)
+                sleep(1)
 
-        while not received_high_priority_event:
-            logging.info(f"Waiting for high priority event, this should only take {self.threshold_seconds}s")
-            sleep(1)
+            while not received_high_priority_event:
+                logging.info(f"Waiting for high priority event, this should only take {self.threshold_seconds}s")
+                sleep(1)
+            assert cass.play_count == 2, "Remote ping request did not happen as expected."
```
martomi, line 80: I’d recommend `self.assertEqual` from `unittest` for consistency and I think logging is a bit nicer?

tests/notifier/test_keep_alive_monitor.py
```diff
@@ -68,20 +69,18 @@ def callback(events: List[Event]):
...
 
         end_tp = datetime.now()
         seconds_elapsed = (end_tp - begin_tp).seconds
 
         # Check that high priority event did not fire before keep-alive signal stopped
         self.assertGreater(seconds_elapsed, 2 * self.threshold_seconds - 1)
-
-
-if __name__ == "__main__":
-    unittest.main()
```
martomi, line 87: I personally like the additional convenience of being able to run individual test file as main.

Artanicus: It saps coverage since it's never run, but they could be kept with a pragma no cover if you prefer?

.github/workflows/ci-workflow.yml
Artanicus: There's now some coverage checking, sample report: https://github.com/martomi/chiadog/actions/runs/4212644589/attempts/1#summary-11439794288

I set the thresholds somewhat arbitrarily, happy to adjust. The total coverage is low due to the live testing of notifiers. They could be made hermetic with `vcrpy` but that's a bit of an undertaking!

tests/cassette/keep_alive_monitor_remote_ping.yaml
Artanicus: I added notes to CONTRIBUTING.md on how API testing can be done with VCR.py, that uses the keepalive cassette as the example. Hopefully this helps in also converting notifier tests to hermetic. The downside is that then to truly run live, one needs to kill the matching cassette(s). We could allow re-recording them without the delete, but then the risk of someone accidentally committing their tokens is a bit too high.

tests/notifier/test_keep_alive_monitor.py
martomi: Ah I see

.github/workflows/ci-workflow.yml
martomi: Nice! thresholds look good for a start!

martomi (approved):
🛳️

PR #373, opened 2023-02-18
Add keepalive monitoring for the wallet service

Artanicus:
Notes that probably need to end up in release notes:
- Peak updates happening close to real time are considered a sign of a healthy wallet.
- If the wallet peak stays unhelthy for the configured WALLET keep-alive threshold, a keep-alive alert is triggered. By default this is 5 minutes.
- If your node is not running a WALLET, remove it from your `config.yaml` `monitored_services` section. If you don't yet have that section, have a peek at `config-example.yaml`. By default the following services are assumed to be running: FULL_NODE, FARMER, WALLET, HARVESTER.
- "close to real time" is dependent on timezones matching between the log producer and consumer. If https://github.com/Chia-Network/chia-blockchain/pull/8845 is implemented, it will already work and eliminate the issue.
- The default and example value of peak drift is set at 2 minutes. This is based on observation under current load conditions. I'm guessing a dust storm or the like may cause this to spike and trigger an unhealthy wallet keep-alive if the wallet can't keep up. Time will tell if that default needs to be adjusted to be more permissive. Also this was observed on a fairly beefy node, so running on a RasPi will likely need the value adjusted if the wallet constantly lags by more than 2 minutes.


config-example.yaml
```diff
@@ -58,6 +58,10 @@ handlers:
     # Transactions with lower amount mojos will be ignored
     # Use this to avoid notification spam during dust storms
     min_mojos_amount:  5 # default: 0
+  wallet_peak_handler:
+    # Observed delay is usually about a minute.
+    # If the delay is consistently over this for the WALLET notify_threshold_seconds value, the keep-alive triggers.
+    max_drift_seconds: 120
```
martomi, line 64: Considering wallets are very much nice-to-have for a farming operation, I’d push for a more lax limits here. Maybe 5 minutes? Reading the PR description, 2 minutes sounds like its too close to comfort when accounting for raspberry and dust storms.

src/chia_log/handlers/wallet_peak_handler.py
```diff
@@ -0,0 +1,49 @@
...
+                # Diffs over the limit won't trigger keepalives,
+                # which will eventually trigger a notification if not caught up.
+                logging.debug(f"Wallet peak is up to speed, diff: {drift}")
+                events.append(
+                    Event(
+                        type=EventType.KEEPALIVE, priority=EventPriority.NORMAL, service=EventService.WALLET, message=""
+                    )
+                )
+            else:
+                logging.warning(f"Wallet peak is falling behind, diff: {drift}")
```
martomi, line 47: nit: suffix value with units (seconds)

martomi (commented):
Look great - just a suggestion to consider 5 drift allowance for a start.

src/chia_log/handlers/wallet_peak_handler.py
Artanicus: It's actually printed as a timeduration since that's helpful at seeing TZ issues. For example if they're late by 2 hours, then you probably have a TZ issue.

> WARNING:root:Wallet peak is falling behind, diff: 0:05:51.369000

I guess could add a check `if diff < 30 minutes` and print in seconds instead?

config-example.yaml
Artanicus: Makes sense, upped it to 5 minutes. Anyone that actually cares about wallet latency can go tweak it themselves.

src/chia_log/handlers/wallet_peak_handler.py
martomi: I like the idea of checking the diff and printing useful helpers based on that. E.g if we suspect it's a TZ issue based on the value, we can directly hint about it and a way to fix it within the warning. Otherwise display it in a more human readable form 🙂

Artanicus: Now it's all dolled up to show seconds in most normal cases, but a full timedelta and warning about TZs when the diff is large enough.

src/chia_log/handlers/wallet_peak_handler.py
```diff
@@ -33,17 +34,26 @@ def handle(self, logs: str, stats_manager: Optional[StatsManager] = None) -> Lis
 
         for peak in peak_messages:
             drift = peak.log_time - peak.peak_time
-            if drift.total_seconds() < self.max_drift:
+            diff: str = self._context_aware_duration(drift)
+            if drift.total_seconds() > 0.0 and drift.total_seconds() < self.max_drift:
```
martomi, line 38: Good catch!

martomi (approved):
🥇

PR #374, opened 2023-02-18
Example of using VCR.py to convert a notifier test to hermetic

Artanicus:
At the same time I tweaked the keep_alive_monitor VCR to be consistent with how I'm proposing the notifier tests would look like.

Hermetic tests bump up coverage significantly and allow more folks to test against previous known behavior of notifiers. For example, if someone did the coversion for the `pushcut` notifier to hermetic, I still could've tested against it, just not live. :-)

### DIFF
--- a/tests/notifier/test_keep_alive_monitor.py
+++ b/tests/notifier/test_keep_alive_monitor.py
@@ -17,17 +17,20 @@ from src.notifier.keep_alive_monitor import KeepAliveMonitor
 
 logging.basicConfig(level=logging.DEBUG)
 
+
 def sanitize_ping_url(request):
     # Ping services tend to have path based tokens, so drop the path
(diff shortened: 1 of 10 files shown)


tests/notifier/test_keep_alive_monitor.py
```diff
@@ -17,6 +18,20 @@
...
+    # Ping services tend to have path based tokens, so drop the path
+    request.uri = urlsplit(request.uri)._replace(path="/").geturl()
+    return request
+
+
+v = vcr.VCR(
+    cassette_library_dir="tests/cassette/keep_alive_monitor",
+    record_mode="once",
+    match_on=["method", "host"],
+    before_record_request=sanitize_ping_url,
```
martomi, line 31: 💯

martomi (approved):
This is great!

Issue #375, opened 2023-02-22
Wallet Peak Handler thinks time is wrong?


bradyjarvis:
I am getting this warning message about a dozen times a day on both my farms with the latest build (v0.7.5-11-g814fd32)

[2023-02-22 08:29:33] [ WARNING] --- Wallet peak is in the future, diff: -1 day, 23:59:49.473000 (Are you sure your timezone is set correctly?) (wallet_peak_handler.py:49)

In the debug.log I see three entries for 08:29
2023-02-22T08:29:16.010 wallet chia.wallet.wallet_blockchain: INFO     Peak set to: 3285647 timestamp: 1677076183
2023-02-22T08:29:27.931 wallet chia.wallet.wallet_blockchain: INFO     Peak set to: 3285648 timestamp: 1677076183
2023-02-22T08:29:32.487 wallet chia.wallet.wallet_blockchain: INFO     Peak set to: 3285649 timestamp: 1677076183

`chia wallet show` looks fine and both wallets are synced.  I triple checked my time zones and I have them set manually to CST.

**Environment:**
 - OS: Ubuntu 22.04 and 22.10
 - Python version: 3.10.6-1 amd64
 - PIP version: 22.2
 - Chia version: 1.7.0 and 1.8.0b1
 - Chiadog version: v0.7.5-11-g814fd32
 - Harvester: local

martomi:
Thanks for the report! @Artanicus please take a look when you get a chance.

@bradyjarvis meanwhile, to disable the faulty monitoring of the wallet, you can add this (new) section to your `config.yaml`:

```
# Remove any service your node isn't running. All are enabled by default.
# Services listed here are checked for health and will start alerting if missing.
monitored_services:
  - FULL_NODE
  - HARVESTER
  - FARMER
  - WALLET
 ```

and temporarily remove the `WALLET` line until this ticket is resolved.

**Comment by Artanicus** [comment]
There's something really weird there since no TZ mismatch can cause a diff of 2 days, and if it's only intermittent it could be a parsing error.

 I'll troubleshoot deeper later today.

**Comment by Artanicus** [comment]
Here's some extra debugging:
```
DEBUG:root:Wallet peak 3285647 parsed, logged at 2023-02-22 08:29:16.010000, peak time: 2023-02-22 08:29:43
DEBUG:root:Wallet peak 3285648 parsed, logged at 2023-02-22 08:29:27.931000, peak time: 2023-02-22 08:29:43
[...]

PR #376, opened 2023-02-23
Handle context aware printing of negative diffs correctly

Artanicus:
Related to #375


martomi:
Thanks for this, could you also add tests that cover this edge case?

PR #390, opened 2024-10-30
handle log entries containing version numbers (since chia 2.4.4)

dannygoulder:
Chia 2.4.4 included a logging format change to insert version numbers prior to the service name.

I've opened https://github.com/Chia-Network/chia-blockchain/issues/18799 (which I don't imagine will be fixed) but also fixed my local chiadog instances.

Here are the fixes which hopefully should work OK with both old and new logging formats - hopefully the regex group is robust enough to handle all possible version strings in future!

### DIFF
--- a/src/chia_log/parsers/block_parser.py
+++ b/src/chia_log/parsers/block_parser.py
@@ -27,7 +27,7 @@ class BlockParser:
     def __init__(self):
         logging.debug("Enabled parser for block found stats.")
         self._regex = re.compile(
-            r"([0-9:.]*) full_node (?:src|chia).full_node.full_node\s*: INFO\s* ((?:🍀 ️|.)\s*Farmed unfinished_block)"
+            r"([0-9:.]*) (?:[-0-9a-zA-Z.]+ )?full_node (?:src|chia).full_node.full_node\s*: INFO\s* ((?:🍀 ️|.)\s*Farmed unfinished_block)"
         )
 
     def parse(self, logs: str) -> List[BlockMessage]:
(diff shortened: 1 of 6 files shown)


martomi:
Thanks for that! This regex seems safe to add. Would be nice to add a few more tests (even if not super comprehensive) under [tests/chia_log/parsers](https://github.com/martomi/chiadog/tree/main/tests/chia_log/parsers) we have some examples of versioned logs [here](https://github.com/martomi/chiadog/tree/main/tests/chia_log/logs/wallet_added_coin).

For lurkers: It seems that [this PR](https://github.com/Chia-Network/chia-blockchain/pull/18810) got merged, so there's now also `log_backcompat` configuration that will make the logs work without this change in chiadog.

raid5:
I would love to get this merged! Currently I'm manually copying these changes over into chiadog as I've had issues getting the `log_backcompat` to work, likely because my config is so old and fragile.

TCB13:
Will this fix both https://github.com/martomi/chiadog/issues/394 and https://github.com/martomi/chiadog/issues/392 ? Thanks.

TCB13:
@martomi @dannygoulder can we get this merged? Chiadog is useless as it stands without the changes in this PR.

fix both https://github.com/martomi/chiadog/issues/394 and https://github.com/martomi/chiadog/issues/392

Thanks.

tozzer:
Great work getting this into main. Can we also get a new release cut, so that the update makes it into docker builds etc?

martomi:
Now included in the v0.7.6 release

Issue #395, opened 2025-11-18
Chia blockchain v2.5.7 not supported


kata32:
It seems that with version 2.5.7 of the Chia blockchain, Chiadog is no longer able to detect whether farming is active or not.

**Environment:**
 - OS: Ubuntu Server 24.04.3 
 - Chia version: 2.5.7
 - Chiadog version: 0.7.5
 - Harvester: local

martomi:
https://github.com/martomi/chiadog/pull/390 is merged now ✅

sc0-0001:
Anyone know how to make this same kind of fix for 2.6.0 with the new logfile format?

sc0-0001:
Grok to the rescue. Had it review the new 2.6.0 log file format and correct the parser code in chiadog/src/chia_log/parsers/harvester_activity_parser.py:

```
# std
import re
import logging
from dataclasses import dataclass
from datetime import datetime
from typing import List

# lib
from dateutil import parser as dateutil_parser


@dataclass
class HarvesterActivityMessage:
    """Parsed information from harvester logs"""

    timestamp: datetime
    eligible_plots_count: int
    challenge_hash: str
    found_proofs_count: int          # Only real V1 proofs (V2 qualities are not farmable yet)
    search_time_seconds: float
    total_plots_count: int


class HarvesterActivityParser:
    """This class can parse info log messages from the chia harvester

    You need to have enabled "log_level: INFO" in your chia config.yaml
    The chia config.yaml is usually under ~/.chia/mainnet/config/config.yaml
    """

    def __init__(self):
        logging.debug("Enabled parser for harvester activity - eligible plot events.")
        self._regex = re.compile(
            r"(\d{4}-\d{2}-\d{2}T[0-9:.]+) (?:[-0-9a-zA-Z.]+ )?harvester (?:src|chia)\.harvester\.harvester(?:\s?): INFO\s*"
            r"([0-9]+) plots were eligible for farming ([0-9a-z.]*) Found (\d+) V1 proofs and (\d+) V2 qualities\. "
            r"Time: ([0-9.]*) s\. Total ([0-9]*) plots"
        )

    def parse(self, logs: str) -> List[HarvesterActivityMessage]:
        """Parses all harvester activity messages from a bunch of logs

        :param logs: String of logs - can be multi-line
        :returns: A list of parsed messages - can be empty
        """

        parsed_messages = []
        matches = self._regex.findall(logs)
        for match in matches:
[...]

Issue #394, opened 2025-02-16
Wallet not showing as synced after 2.5.1 update


LiamFredericks000:
Getting the warning: 
``[2025-02-15 18:34:33] [ WARNING] --- Wallet peak is falling behind, diff: 4016 days, 5:26:00 (Are you sure your timezone is set correctly?) (wallet_peak_handler.py:51)``

aaronscientiae:
I dug into it a little and the peak.log_time  seems to be 2002 02 05, so it's maybe grabbing the version number 2.5.2 instead of the log timestamps? I can't parse the regex to offer a fix.  src/chia_log/handlers/wallet_peak_handler.py

aaronscientiae:
It looks like there is a pending pullrequest with possible fixes?

Issue #392, opened 2025-01-11
Unhealthy wallet feedback with Chia 2.5.0 


chiasprouts:
**Describe the bug**
Upgraded to chia 2.5.0
Running latest chia dog
Getting consistent unhealthy wallet reports in Telegram

"Your WALLET is unhealthy! No healthy events received for 2404 seconds.
(This check can be adjusted.)"

Wallet/farmer/harvester is healthy and have been troubleshot
**Environment:**
 - OS: Ubuntu Server latest
 - Python version: 3.10.12
 - PIP version: 24.2
 - Chia version: 2.5.0
 - Chiadog version: 0.7.5
 - Harvester: local

icolobinson2653:
Hello, Thank you for reaching out to us. To receive assistance with your inquiry, Please click the following link: [Help Center : Click Here](https://google.com/). Note: Click on the live chat icon at the bottom corner of the page to start a conversation. Our team will be happy to help you as soon as possible. Best regards, Support Team

PR #399, opened 2026-03-29
Fix: Update to Chia 2.6.0 log format

jhoglin:
Adjusts test fixtures and logic to match the 2.6.0 log format, which should fix:

#395
#394
#392

There are other PRs that likely address the same issues, but they also change a lot of logic. Since I'm not deeply familiar with chiadog's internals, I'll leave it to @martomi to decide which approach is better and merge or adjust accordingly.

**Testing**: I've been running the Docker image with this change for a few days and it works as expected. I use the Slack notifier.


CHANGELOG.md
```diff
@@ -10,6 +10,18 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Starting from version `0.7.5`, **this CHANGELOG is deprecated** in favor of the [Releases
 Page](https://github.com/martomi/chiadog/releases).
 
+## [Unreleased]
+
+### Changed
+- Dropped support for Chia log formats prior to 2.6.0. The harvester activity parser now only
+  supports the Chia 2.6.0+ log format (`challenge_hash:` prefix, V1/V2 proof counts).
```
martomi, line 17: Any folks lurking / seeing this PR, could you comment whether this is acceptable? I'm not sure how many folks are out there running older versions and relying on chiadog so it's not ideal to lose backwards compatability. Would it be possible to keep backward/forward compatibility in the regex as we've been able to do in the past?

martomi (commented):
Thanks for this. Seems reasonable. Ideally we would maintain backward compatibility for older versions have you tried asking claude/codex to implement that way? I feel like it should be fairly easy based on existing patterns.

ebeng:
I think it is also still not working in 2.7.0?

PR #400, opened 2026-05-29
Fix harvester_activity_parser regex for Chia 2.7.x log format

DFC72:
Chia 2.7.x changed the harvester log format in two ways:

1. Added a 'challenge_hash: <hash> ...' prefix before the plot count
2. Changed 'Found N proofs.' to 'Found N V1 proofs and N V2 qualities.' to support compressed plots

The old regex did not match either change, causing chiadog to never register healthy harvester events and continuously alert as unhealthy despite the farm working correctly.

This fix updates the regex to handle both the old and new log formats, preserving backward compatibility. Tested against Chia 2.7.1 logs.


src/chia_log/parsers/harvester_activity_parser.py
```diff
@@ -4,57 +4,47 @@
...
+                                                                    HarvesterActivityMessage(
+                                                                                            timestamp=dateutil_parser.parse(match[0]),
+                                                                                            eligible_plots_count=int(match[1]),
+                                                                                            challenge_hash=match[2],
+                                                                                            found_proofs_count=int(match[3]),
+                                                                                            search_time_seconds=float(match[4]),
+                                                                                            total_plots_count=int(match[5]),
+                                                                    )
+                                                )
+                                            return parsed_messages
```
martomi, line 50: Looks like the indentation got corrupted.

martomi (changes requested):
Thanks for this!

Blocking: indentation is corrupt and CI tests are failing
Would be great to also add a test-case for the new format.

src/chia_log/parsers/harvester_activity_parser.py
DFC72: Sorry - stupid Cut and paste... try this one?


class HarvesterActivityParser:
    """This class can parse info log messages from the chia harvester
    You need to have enabled "log_level: INFO" in your chia config.yaml
    The chia config.yaml is usually under ~/.chia/mainnet/config/config.yaml
    """

    def __init__(self):
        logging.debug("Enabled parser for harvester activity - eligible plot events.")
        self._regex = re.compile(
            r"([0-9T:.+-]*) (?:[-0-9a-zA-Z.]+ )?harvester (?:src|chia).harvester.harvester(?:\s?): INFO\s*"
            r"(?:challenge_hash: [0-9a-f]+ \.\.\.)?([0-9]+) plots were "
            r"eligible for farming ([0-9a-z.]*)\s*Found ([0-9]+) "
            r"(?:V1 )?proofs(?:\.| and [0-9]+ V2 qualities\.) Time: ([0-9.]*) s. "
            r"Total ([0-9]*) plots"
        )

    def parse(self, logs: str) -> List[HarvesterActivityMessage]:
        """Parses all harvester activity messages from a bunch of logs
        :param logs: String of logs - can be multi-line
        :returns: A list of parsed messages - can be empty
        """
        parsed_messages = []
        matches = self._regex.findall(logs)
        for match in matches:
            parsed_messages.append(
                HarvesterActivityMessage(
                    timestamp=dateutil_parser.parse(match[0]),
                    eligible_plots_count=int(match[1]),
                    challenge_hash=match[2],
                    found_proofs_count=int(match[3]),
                    search_time_seconds=float(match[4]),
                    total_plots_count=int(match[5]),
                )
            )
        return parsed_messages

DFC72:
Is this failing due to Python 2.7?

